PR #102, opened 2014-08-08
handling unescaped templates and new 'new' target

andrzejsliwa:
Hi,

I made unescaped version of templates, to simplify adding new templates without boilerplate escaping.

new version: 

```makefile
define bs_appsrc
{application, $(PROJECT), [
	{description, ""},
	{vsn, "0.1.0"},
	{modules, []},
	{registered, []},
	{applications, [
		kernel,
		stdlib
	]},
	{mod, {$(PROJECT)_app, []}},
	{env, []}
]}.
endef
```

old version:
```makefile
bs_appsrc = "{application, $(PROJECT), [" \
	"	{description, \"\"}," \
	"	{vsn, \"0.1.0\"}," \
	"	{modules, []}," \
	"	{registered, []}," \
	"	{applications, [" \
	"		kernel," \
	"		stdlib" \
	"	]}," \
	"	{mod, {$(PROJECT)_app, []}}," \
	"	{env, []}" \
	"]}."
```
I modified also **new** target to handle TEMPLATE and NAME directly (without named params):

```bash
$ make new gen_server my_gen_srv
``` 

vs old:

```bash
$ make new t=gen_server n=my_gen_srv
```

the cons of change is that new target can't be chained, but I think this not a case for it.

what do you thinking guys about such improvement?

Cheers,
Andrzej

### DIFF
--- a/erlang.mk
+++ b/erlang.mk
@@ -256,233 +256,283 @@ help::
 
 # Bootstrap templates.
 
-bs_appsrc = "{application, $(PROJECT), [" \
-	"	{description, \"\"}," \
-	"	{vsn, \"0.1.0\"}," \
-	"	{modules, []}," \
-	"	{registered, []}," \
-	"	{applications, [" \
-	"		kernel," \
-	"		stdlib" \
-	"	]}," \
-	"	{mod, {$(PROJECT)_app, []}}," \
-	"	{env, []}" \
-	"]}."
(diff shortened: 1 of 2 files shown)


essen:
I actually did just merge that. I changed the way the echo was done since it didn't work for me. I believe this solution should work for everyone.

This is xmas!

essen:
Thanks!

PR #165, opened 2014-11-14
Support dependencies fetched as git submodules

danielwhite:
This adds an alternative argument, 'submodule', as an alternative to a
URL for git dependencies. For example:

dep_cowboy = git submodule

This option is not propagated to sub-dependencies and all required
submodules should be added to the root repository.

The main benefit is that in later version of git, all references are
stored in the main repositories directory (i.e. .git/modules). This
allows for speedy clean and build without remote cloning every time.

### DIFF
--- a/core/deps.mk
+++ b/core/deps.mk
@@ -43,7 +43,9 @@ distclean:: distclean-deps distclean-pkg
 # Deps related targets.
 
 define dep_fetch
-	if [ "$$$$VS" = "git" ]; then \
+	if [ "$$$$VS" = "git" -a "$$$$REPO" = "submodule" ]; then \
+		git submodule update --init -- $(DEPS_DIR)/$(1); \
+	elif [ "$$$$VS" = "git" ]; then \
 		git clone -n -- $$$$REPO $(DEPS_DIR)/$(1); \
 		cd $(DEPS_DIR)/$(1) && git checkout -q $$$$COMMIT; \
 	elif [ "$$$$VS" = "hg" ]; then \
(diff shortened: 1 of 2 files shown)


danielwhite:
Sorry I've taken so long to get back to this.  I've rebased against the latest, and fixed my example so it's actually based on cowboy.

```
$ git clone --recursive -b submodule-example https://github.com/danielwhite/cowboy.git
```

The problem you see is due to git's default clone creating a directory each submodule. This makes it somewhat difficult to detect the state of the submodule without teaching erlang.mk more about the workings.

Frankly, this is probably too complicated, and can be easily addressed by cloning with the `--recursive` flag.  Perhaps it would be enough to include this caveat in the documentation?

essen:
One thing I wonder, does submodule work properly with erlang.mk? In particular, "distclean" does an "rm -rf deps".

If not, then I would advise just doing --recursive indeed. I will try to add this week the LIB_DIR functionality I mentioned in another ticket, which is like a permanent "deps" dir (think rebar "apps" except the base directory must be an erlang app), and your submodules would fit better in that directory imo.

danielwhite:
With `distclean`, this works perfectly fine because the `deps/{dep}` rule is matched and `git submodule` is invoked to check out the appropriate commit.

`--recursive` is only ever needed when first cloning.

essen:
OK then I'm not sure I understand, is the patch still needed? I see no problem merging the PR as it is.

danielwhite:
I think it's still useful, and haven't had any problems with `distclean` since originally introducing it in my repositories.

essen:
Used your work to add the functionality, with a few changes: you need to say `dep_cowboy = git-submodule`, I made it a separate fetch method entirely.

Thanks!

PR #167, opened 2014-11-16
Make it possible to create erlydtl beam files with full path in the name

brigadier:
if DTL_FULL_PATH=true
  /templates/a/b/c.dtl -> a_b_c_dtl.beam
otherwise (as it was before)
  -> c_dtl.beam

With this change it will be possible to create templates with the same name in different directories.


brigadier:
k, amended the commit message

essen:
Merged, thanks!

PR #178, opened 2014-12-11
Recompile on hrl changes without touching source files

gar1t:
The prior behavior modified source files (using touch) to trigger
a recompile whenever a hrl file was changed. This confuses editors,
which think the source files were actually changed by someone else.

The new approach set the date of the *.app file back to 00/01/01,
which has the same effect without confusing editors.

### DIFF
--- a/erlang.mk
+++ b/erlang.mk
@@ -290,7 +290,9 @@ clean:: clean-app
 
 erlc-include:
 	-@if [ -d ebin/ ]; then \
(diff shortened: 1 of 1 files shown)


essen:
The current version of Erlang.mk will recompile only the source files that have been changed. Unfortunately, this requires us to perform a touch of .erl files. Fortunately, it'll only be for the files that need recompiling, so it won't be as confusing as it was before.

I will document this as a limitation and close the ticket. Thanks!

PR #186, opened 2014-12-21
Tests for erlang.mk targets

zuiderkwast:
Makefile-based tests for a few of the erlang.mk targets. Consider this a proof of concept. It should probably be split up into one makefile per plugin or something like that.

There are several verbosity levels (V=0, 1, 2, 3) for various output. Not sure if all of these are useful.

This is the output of `make tests`:

```
make -C tests
Setting up app.
app: Testing the 'app' target.
Checking the modules line in the generated .app file.
Checking that 'make clean-app' deletes ebin.
Checking that 'make app' returns non-zero on compile errors.
Test 'app' passed.
ct: Testing tests-ct and related targets.
Setting up test suite.
Checking files created by 'make tests-ct'.
Checking that 'make clean-ct' does not delete logs.
Testing target 'ct-mysuite' where mysuite_SUITE is a test suite.
Checking that 'make tests-ct' returns non-zero for a failing suite.
Checking that 'make ct-distclean' deletes logs.
Cleaning up test data.
Test 'ct' passed.
+---------------------+
|  All tests passed.  |
+---------------------+
```

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -17,7 +17,7 @@ BUILD_CONFIG = $(shell sed "s/\#.*//" $(BUILD_CONFIG_FILE))
 
 ERLANG_MK = erlang.mk
 
-.PHONY: all
+.PHONY: all check
 
 all: pkg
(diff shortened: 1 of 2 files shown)


essen:
Merged, thanks!

PR #187, opened 2014-12-21
Update version when building erlang.mk.

kaos:
For me, I like to know which version I have, when running `make help` for instance.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -16,11 +16,13 @@ BUILD_CONFIG_FILE ?= $(CURDIR)/build.config
 BUILD_CONFIG = $(shell sed "s/\#.*//" $(BUILD_CONFIG_FILE))
 
 ERLANG_MK = erlang.mk
+ERLANG_MK_VERSION = $(shell git describe --tags --dirty)
 
 .PHONY: all
 
 all: pkg
-	awk 'FNR==1 && NR!=1{print ""}1' $(patsubst %,%.mk,$(BUILD_CONFIG)) > $(ERLANG_MK)
+	awk 'FNR==1 && NR!=1{print ""}1' $(patsubst %,%.mk,$(BUILD_CONFIG)) \
+	| sed 's/^ERLANG_MK_VERSION = .*/ERLANG_MK_VERSION = $(ERLANG_MK_VERSION)/' > $(ERLANG_MK)
 
 pkg:
 	cat packages.v2.tsv | awk 'BEGIN { FS = "\t" }; { print $$1 "\t" $$3 "\t" $$5 "\t" $$6 }' > packages.v1.tsv
--- a/erlang.mk
+++ b/erlang.mk
@@ -14,7 +14,7 @@
 
 .PHONY: all deps app rel docs tests clean distclean help erlang-mk
 
-ERLANG_MK_VERSION = 1
+ERLANG_MK_VERSION = 1.2.0-37-g5fef667-dirty
 
 # Core configuration.
 


essen:
Hey, if you're around can you squash the 2 commits into one?

kaos:
Sure.

kaos:
Btw, I just now realized that since you commit the `erlang.mk` file as well.. that you will always get a version pointing at the parent commit. And if your commit also includes other changes and not only the erlang.mk file, then the version will be misleading. So you want to make two commits. One for the fuctionality, and one for the generated `erlang.mk` file, which then will refer to sensible commit.

This is perhaps not really what you want, and one way to mitigate the issue is to also use the `--dirty` flag on the version string. That way you will at least know when the commit referenced in the version string has some additional changes done to it as well (i.e. a rebuilt `erlang.mk` committed along with the source changes).

Example version, using dirty flag:
```
1.2.0-37-g5fef667-dirty
```

kaos:
I pushed another update with the `--dirty` flag included.
It's easy to pull out if you don't like it. :)

essen:
I don't think it helps (so you can remove it I guess) because the dirty will be applied not only to the erlang.mk committed here, but also to the ones people do manually, so the version will be wrong in half the cases. The solution to this would be to add a commit hook in addition to what you did. So we want to duplicate this change in a `pre-commit` file. Something like http://higherabstractionlevel.com/git-pre-commit-hook-to-put-revision-number-in-source-file/ but suited to our needs (and no need for a Python script of course). I'm not sure one can PR a pre-commit file but either way I will look into it in more details tomorrow.

essen:
Unfortunately hooks are not cloned with a repository so that's not a solution.

essen:
Actually I could very well use the post-merge hook to update the git number after a merge. So that would be enough. Nobody else would need hooks, just me. OK then. As for dirty, yeah, let's leave it. I will work on those hooks and merge tomorrow.

essen:
Hey, actually we can't. The git number is only available once the commit is done, and trying to amend or similar results in a new commit number. Not what we want. Also here: http://stackoverflow.com/questions/4835035/can-a-git-pre-commit-hook-add-a-file-to-the-repo?rq=1

For future reference this is what my hook looked like:

``` bash
#!/bin/sh

if [ ! -f erlang.mk.tmp ]
then
	ERLANG_MK_VERSION=$(git describe --tags --dirty)
	cp erlang.mk erlang.mk.tmp
	cat erlang.mk.tmp | sed "s/^ERLANG_MK_VERSION = .*/ERLANG_MK_VERSION = $ERLANG_MK_VERSION/" > erlang.mk
	git commit --amend --no-edit erlang.mk
	rm erlang.mk.tmp
fi
```

Doesn't look like there's anything we can do about it. If you got an idea...

essen:
I will close this. I don't think this solution can ever work. Thanks for trying!

kaos:
Thanks for considering it. I'll re-open in case I find a way to get this to work.. :p

kaos:
essen,

I was just thinking.. why do you commit the `erlang.mk` file at all?

When I want to update erlang.mk, I run `make erlang-mk` which git clones into a temp build dir and rebuilds it for me, so then that was committed is discarded any way.

So, the only reason for its existence as far as I can tell is for bootstrapping. And for that reason the file could live somewhere else, right? (i.e. on a github wiki page).

The breaking point in this PR was the fact that we couldn't commit a proper version in erlang.mk.. but if we never commit erlang.mk, that issue is void.

What do you think?

essen:
I commit it to make it easy for people to get it, but I suppose now that the erlang-mk target exists, having a small erlang.mk file that simply bootstraps is a possibility.

I'm not sure what the best solution would be, but I know we want the file that bootstraps in the repository and not lost in a github page somewhere.

kaos:
Bah.. it keeps the old references around..

I couldn't make a PR for it, as it's a new branch, but you could make your own.. or use [mine](https://github.com/kaos/erlang.mk/tree/bootstrap) to see what it looks like. Then add a link to it in the README on the master branch. That way it's easy to update too ;)

essen:
Not sure what you mean?

kaos:
To have the bootstrap version of erlang.mk in the repo, I suggested to have it in it's own branch (the one I linked to) that doesn't have any other history than for itself.

I couldn't make a PR for it, as it is a new branch, so github doesn't deal with that, apparently.
But, you could set it up yourself just as easily:
```
git checkout --orphan bootstrap
git rm -rf .
cp ../stashed/erlang.mk erlang.mk
edit README.md
git add .
git commit && git push -u origin bootstrap
```

The benefit being that it's easy to update the bootstrapped version of erlang.mk simply by pushing a new version to the bootstrap branch.

essen:
Not what I meant.

I think it's OK to not have the complete file anymore. However I think the file people download should be at the same URL as before.

It's OK if the first time erlang.mk is ran it performs the extra bootstrap setup. Perhaps it's even possible to get the erlang.mk bootstrap to update the erlang.mk file and then call 'make' on the new file using the same arguments as before, so in essence it will just make the first invocation a little slower.

The perfect solution is that people still do as they do today, get the file from here and use it, and that erlang.mk takes care of the extra steps required for the bootstrap.

kaos:
Ah.. now I finally got your initial meaning for a bootstrap file, after reading it a few more times.. :p 

No need for a bootstrap branch then.. ;) but it's slightly more work, so I don't think I'll have the time to make a PR for it (not right now, any how).

essen:
Let's reopen and leave it at that for now then so we don't forget. I think it would be a pretty cool addition but don't have much time right now either.

Cheers.

essen:
Now that the bootstrap is merged perhaps we can add the version change when the bootstrap or 'make erlang-mk' runs?

kaos:
:+1: :)
 Den 9 maj 2015 18:33 skrev "Loïc Hoguin" <notifications@github.com>:

essen:
I was gonna say "you can do it" but it looks like the current patch is actually good.

essen:
Merged, thanks!

kaos:
Thank you :)

PR #189, opened 2014-12-23
Build deps, app, rel targets really in that order

stolen:
This pull requests adds the power of [order-only prerequisites](https://www.gnu.org/software/make/manual/html_node/Prerequisite-Types.html) feature of GNU make to erlang.mk.
The code is built really after its dependencies.
This fixes the need of multiple make runs with ```MAKEFLAGS=-j32```


essen:
Wooooo I have to try this.

essen:
Wooooo merged! Also documented in readme and make help. Thanks!

PR #192, opened 2014-12-26
Allowing skipping deps target

grahamrhay:
Same effect as rebar's "skip_deps=true". Particularly handy when running the eunit task (~1s vs ~6s, on my machine). There's probably a better way to accomplish this, but I'd really like the feature :)

### DIFF
--- a/core/deps.mk
+++ b/core/deps.mk
@@ -30,6 +30,8 @@ PKG_FILE_URL ?= https://raw.githubusercontent.com/ninenines/erlang.mk/master/pac
 # Core targets.
 
 deps:: $(ALL_DEPS_DIRS)
+ifeq ($(SKIP_DEPS),true)
+else
 	@for dep in $(ALL_DEPS_DIRS) ; do \
 		if [ -f $$dep/GNUmakefile ] || [ -f $$dep/makefile ] || [ -f $$dep/Makefile ] ; then \
 			$(MAKE) -C $$dep ; \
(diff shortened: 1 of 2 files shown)


essen:
I am considering adding this anyway regardless of `AUTOPATCH`.

grahamrhay:
I'm happy to keep it as a fork, but it's definitely something I find useful for now. However, if the autopatch feature makes the time for the deps target negligible, then I'll have no need for it.

essen:
Hey, the autopatch feature is in. Still want this in though, but not convinced on how it should be done. Will get back to you.

grahamrhay:
I'll give it a try, thanks

kaos:
There's an easier way to accomplish this.

Simply define your deps conditionally in the `Makefile`, like so:
```
DEPS ?= foo bar baz
...
include erlang.mk
```

Then when you invoke:
```
DEPS= make tests
```

it will run the tests, skipping all deps. No change to `erlang.mk` needed. ;)

grahamrhay:
good idea, thanks!

essen:
Please keep this open I want to add this trick to the documentation. :-)

essen:
I added a SKIP_DEPS variable. Main reason for doing so is that there is not just DEPS= but TEST_DEPS= and DOC_DEPS=. Having one variable requires less trickery.

I did it a little differently than in this PR as I also wanted to exclude the fetching.

Enjoy!

grahamrhay:
thanks!

Issue #193, opened 2014-12-29
Dependency is specified as a dependency but is not reachable by the system.


MarkNijhof:
I have two projects, core and app. app depends on core and core depends on lager (this is not lager specific) I created some basic apps to demonstrate the problem.

app: https://github.com/MarkNijhof/app
core: https://github.com/MarkNijhof/core

Running make in core works as expected and running ./run.sh in there starts the app just fine. Running make in app fails to make a release for core with the following error:

```
===> Starting relx build process ...
===> Resolving OTP Applications from directories:
          /home/mark/src/app/deps/core/ebin
          /usr/lib/erlang/lib
===> Resolving available OTP Releases from directories:
          /home/mark/src/app/deps/core/ebin
          /usr/lib/erlang/lib
Failed to solve release:
 Dependency lager is specified as a dependency but is not reachable by the system.

make[1]: *** [relx-rel] Error 127
make[1]: Leaving directory `/home/mark/src/app/deps/core'
make: *** [deps] Error 2
```

After this the app/deps folder does contain a build version of lager.

I can make it work by running make inside app/deps/core

What am I doing wrong?

PR #194, opened 2014-12-29
Updated relx to version v1.1.0

MarkNijhof:
Updated relx to version v1.1.0 and tested with the setup from issue #193


essen:
Merged, thanks!

PR #195, opened 2015-01-01
Allow there be no _SUITE in test/.

jinnipark:
There exist a convention to put eunit test codes under test/ directory.  Better to allow this and suppress warnings that there is no ct_suites specified.


essen:
I'm aware of the issue but that's not exactly what I had in mind. I'll get back to you.

essen:
Hey, this should be fixed in 7ee4e7958acf0f0956bd64023ecb1731b2824b5e which introduced various other improvements. Closing, thanks!

PR #199, opened 2015-01-09
Don't overwrite elvis config if elvis binary is missing

srijan:


essen:
Good point, will test and merge right after I finish my current patch.

essen:
Merged, thanks!

PR #202, opened 2015-01-11
Prevent 'make all' from removing elvis

danielwhite:
Given that 'make all' is the default build, it seems surprising to
delete this without an explicit request for a 'distclean'.

@jfacorro could perhaps shed some light on why this behaviour
was initially included.

### DIFF
--- a/erlang.mk
+++ b/erlang.mk
@@ -943,10 +943,6 @@ help::
 		"Elvis targets:" \
 		"  elvis       Run Elvis using the local elvis.config or download the default otherwise"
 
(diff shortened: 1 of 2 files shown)


jfacorro:
@danielwhite I can't think of a reason why `elvis` should be removed when building a release. This is most probably a mistake on my part when I wrote the plugin. Thanks for reporting it.

essen:
Cool, will merge then.

essen:
And merged, thanks!

Issue #200, opened 2015-01-10
Unify 'tests' target for ct, eunit, triq, ...


zuiderkwast:
I know that you @essen are working on this already but I think it could be nice to have an issue where people can add their ideas. You could assign yourself to it so that people can see what's going on. Here are a few of my ideas:

* Rename `build-ct-deps` to `test-deps` and make `test-deps` a prerequisite of `tests` rather than of `build-ct-suites`.
* ct shouldn't fail if there is no `_SUITE` in `test/` (as noted in #195).
* Coverage report using `cover`, preferably for eunit, ct and other kinds of tests, which would require all these to run in the same vm instance. Pseudo code:
  `cover:enable("src/*.erl"), eunit:run($(EUNIT_MODULES)), ct:run($(CT_SUITES)), triq:check($(TRIQ_MODULES)), $(USER_DEFINED_TEST_CODE), cover:report().`
* Make it possible to hook in other unsupported testing methods e.g. proper or quviq eqc using some makefile variable such as `$(USER_DEFINED_TEST_CODE)` above.

essen:
First is done and will be pushed today (with a twist, there's a test-build target that eunit/ct both depend on, this is to enable parallel testing).

Second point, noted, I will try that now.

Cover is out of scope at this point though I am not forgetting, and still think we should go with https://github.com/ramsay-t/Smother instead of the less powerful cover. But either way it should be pluggable so that people who really want cover can later add it.

I will be adding the triq plugin also today. Not sure what you mean with the makefile variable, you can already add your own testing methods the same way eunit was added. Just add e.g. `tests:: proper` line and write the `proper` target.

The bigger change I introduce here is the concept of a "test build" to avoid rebuilding everything for every test tool. Seems to work great in my testing.

zuiderkwast:
Nice. What I meant by the makefile variable is that I'd like a way to run ct and eunit and e.g. proper tests with a combined coverage report for all these testruns. That's what i tried to illustrate with the pseudo code anyway. If there is another way to run them separately and merge separate the cover reports afterwards, that would be fine too. Now I'll have a look at Smother.

essen:
Right this is preferable if it is possible.

essen:
Another thing I want to do is introduce the "t" variable the same way I use it with triq:

$(t) isn't defined - run everything
$(t) is defined and doesn't contain ":" - run the module $(t)
$(t) is defined and contains ":" - run the test case function $(t)/0

[...]

PR #204, opened 2015-01-24
Combined coverage report for eunit and ct

zuiderkwast:
As discussed in #200. Coverdata is collected for eunit and ct separately and a separate target 'cover-report' combines them and produces a HTML report.

I had to make some minor changes in test.mk, eunit.mk and ct.mk to be able to hook in the cover analysis. For eunit, the tests have to be run by module. If the are run using `{dir, "ebin"}` eunit reloads the modules from the files and the cover compilation is lost.

I also generate a simple index.html summary similar to the one generated by rebar (https://github.com/rebar/rebar/blob/master/src/rebar_cover_utils.erl#L185). It would be nice to add some information on what tests were actually run to produce the coverage report. Maybe even the output of the test runs. Any thoughs on this?

Also triq remains to be added but there are some minor glitches (check that test dir exists, check that triq exists, etc.) that could be done first.


essen:
Looks cool, hope I get time to try it soon. :-)

zuiderkwast:
Yes, I hope so too. :-) I could fix the triq stuff also, better in another branch I guess...

I managed to get covertool working with this branch. Covertool is the tool that converts coverage data from cover to Cobertura XML format that works with Jenkins. I'm posting a Makefile here just for reference. It could be moved to a wiki page or something in the future. Btw I love the autopatch feature! (Covertool has rebar as a dep for totally no reason.)

```make
PROJECT = app1
TEST_DEPS = covertool
dep_covertool = git git@github.com:idubrov/covertool.git master
AUTOPATCH += covertool rebar
include erlang.mk

test-build:: deps/covertool/covertool

deps/covertool/covertool:
	$(gen_verbose) make -C deps/covertool escript

test-with-covertool:
	$(gen_verbose) make tests COVER=1 COVER_REPORT_DIR=
	$(gen_verbose) make all.coverdata
	$(gen_verbose) deps/covertool/covertool -cover all.coverdata \
		-output coverage.xml -src src/
```

erlang.mk
```diff
@@ -1266,3 +1274,140 @@ triq: test-build
...
+ifneq ($(COVER_REPORT_DIR),)
+distclean:: cover-report-clean
+endif
+
+help::
+	@printf "%s\n" "" \
+		"Cover targets:" \
+		"  cover-report  Generate a HTML coverage report from previously collected" \
+		"                cover data." \
+		"  all.coverdata Merge {eunit,ct}.coverdata into one coverdata file." \
```
essen: I do not understand the name 'all.coverdata', isn't cover-merge good enough?

build.config
```diff
@@ -22,3 +22,4 @@ plugins/eunit
 plugins/relx
 plugins/shell
 plugins/triq
+plugins/cover
```
essen: Also alphabetical order here.

erlang.mk
```diff
@@ -1,4 +1,4 @@
-# Copyright (c) 2013-2014, Loïc Hoguin <essen@ninenines.eu>
```
essen: I merged a PR that make the changing of this file not needed anymore.

erlang.mk
zuiderkwast: `all.coverdata` is a filename. Is it better to have a non-filename target name?

build.config
zuiderkwast: I'm not sure it works if it is reordered. It contains lines that hook in to already defined variables using `ifdef` and lines such as `CT_RUN += -cover ct.cover.spec` so those things would have to be redesigned. Any ideas?

essen: Then please add a new comment block explaining this plugin must be after all others.

erlang.mk
essen: Is it something that we are supposed to call ourselves then?

How do you use this plugin, can you document it in the README?

zuiderkwast: Yes, I can document it better in the README. (For the record: Normally you only want the HTML coverage report. The file all.coverage cover file is used internally and is otherwise only interesting if you want to use it for something else such as converting it to another coverage format or use other coverage report genarators, etc.)

essen: Understood. I'll try it out.

essen:
OK it works and I will merge it in a few with the small changes I mentioned here.

Something I wonder: how do I read eunit coverage data?

essen:
Hm actually would be great if the cover.spec and coverdata files would be put inside the test/ directory since that's what they're about.

essen:
Will try to fix it I want to merge.

essen:
I will leave the coverdata as is for now but cover.spec definitely goes to test/.

essen:
I think it would be interesting to add a way to also cover compile some deps (for example Cowboy + Cowlib).

essen:
Well I merged. I will open a ticket with more suggestions.

essen:
#236

Thanks for the work!

essen:
I fixed an issue with the test suite and made a minor change to have more readable code. Using it a little more I like it, good job!

Do you think you will have time to add triq support and other things in the future? I created ticket #236 for keeping ideas in. Also don't worry about documentation, I am writing a complete user guide now anyway.

zuiderkwast:
Ok! It's nice that you like it and have merged. I haven't had time to look at this, but I agree about your comments. Coverdata files probably should be in test/ if not in the cover/ directory.

To answer the question "how do I read eunit coverage data?": Use `cover:import/1` and then use other other functions in the cover module. See the cover-report target for example.

Adding triq support would be trivial and I think similar to eunit coverage. I hope I'll have time to do it. Otherwise I'm sure you can do it easily.

Regarding usage documentation, the primary idea is that you run `make test COVER=1` or set `COVER=1` in the Makefile.

essen:
Yeah sorry I answered my own questions about eunit and usage and forgot to say. Cheers. :-)

erlang.mk
```diff
@@ -1266,3 +1274,140 @@ triq: test-build
...
+
+COVERDATA = $(filter-out all.coverdata,$(wildcard *.coverdata))
+
+.PHONY: coverdata-clean
+coverdata-clean:
+	$(gen_verbose) rm -f *.coverdata ct.cover.spec
+
+# Merge all coverdata files into one.
+all.coverdata: $(COVERDATA)
+	$(gen_verbose) $(ERL) -eval ' \
```
jpuigm: Unfortunately this target doesn't work for me,

```
$ make all.coverdata
 GEN    all.coverdata
{"init terminating in do_boot",{{badmatch,{error,{1,erl_parse,["syntax error before: ","'\\'"]}}},[{init,start_it,1,[]},{init,start_em,1,[]}]}}

Crash dump was written to: erl_crash.dump
init terminating in do_boot ()
make: *** [all.coverdata] Error 1
```

However, it does if I make it one-liner,

```
all.coverdata: $(COVERDATA)
        $(gen_verbose) $(ERL) -eval '$(foreach f,$(COVERDATA),cover:import("$(f)") == ok orelse halt(1),) cover:export("$@"), halt(0).'
```

erlang.mk version used: `1.2.0-515-g7dc8c4f-dirty`
platform used: GNU/Linux

essen: Hm it should probably be converted to a $(call erlang,...) form anyway.

essen: Can you open a new ticket or PR?

jpuigm: Sure, will do.

PR #205, opened 2015-01-27
Change version of Erwa master -> 0.1.0

bwegh:
change the version to pull to release 0.1.0 instead of pulling master


essen:
Hey, updated the version directly in a commit to 0.1.1 to avoid adding 4 extra unnecessary commits. Sorry it took so long. Enjoy!

PR #211, opened 2015-02-25
Support for copying dependencies

jozip:
It's sometimes convenient to depend on something you have locally, rather than having to go through a VS.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,0 +1,3 @@
+*~
+*.swp
+.DS_Store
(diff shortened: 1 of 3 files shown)


essen:
Looks good, probably good to merge as is when I find the time. Thanks!

jozip:
Excellent!

essen:
Cherry-picked the relevant commit. Thanks!

Issue #217, opened 2015-03-22
Autopatch issues


essen:
When an application that has no rebar.config but has a Makefile is in AUTOPATCH, nothing is done. We should instead not test for a Makefile existence and instead do the patch. The patch needs to perform all operations as the full autopatch except there is no rebar.config file to parse.

essen:
Example project: https://github.com/dmitryme/erlang_localtime

PR #219, opened 2015-03-26
always patch deps that are listed in the autopatch var.

kaos:
This will also address issue #217. At least for the given example repo.

I took a slightly different approach to that suggested in #217, by looking for the `rebar` script instead of always patching.

### DIFF
--- a/core/deps.mk
+++ b/core/deps.mk
@@ -49,7 +49,8 @@ define dep_autopatch
 	$(ERL) -eval " \
 DepDir = \"$(DEPS_DIR)/$(1)/\", \
 fun() -> \
-	{ok, Conf} = file:consult(DepDir ++ \"rebar.config\"), \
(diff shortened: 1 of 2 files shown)


essen:
Right but I think if someone puts an application to be autopatched we should really just autopatch it, there's really no reason not to I believe.

kaos:
Yeah, or that :)
 Den 26 mar 2015 14:34 skrev "Loïc Hoguin" <notifications@github.com>:

kaos:
I've updated this PR so it always patches. Also, I needed it to honor the `erl_first_files` option from `rebar.config`, so that's in here too.. ;)

essen:
Do you have an example repository for the erl_first_files?

kaos:
This one, for instance.
https://github.com/knutin/elli

But it may have been another that actually caused me compile time issues..
This one, I think it was: https://github.com/uwiger/parse_trans

essen:
Merged, thanks!

PR #220, opened 2015-03-26
Require makefile to build dep. Fixes #218.

### LINKED ISSUE #218: Compiling dep without Makefile is broken
If a dep is missing a Makefile, it is using `erlang.mk` by default, which is fine for.. oh, 2us, after which it fails complaining that it can't make the `app` target.

```
make[2]: *** No rule to make target `app'.  Stop.
make[1]: *** [all] Error 2
```

The reason is simple enough, as you invoke make with this line:

```
echo "include $(CURDIR)/erlang.mk" | ERLC_OPTS=+debug_info $(MAKE) -f - -C $$dep ; \
```

Which will make the default `all` target, which invokes another instance of make using:

```
@$(MAKE) --no-print-directory app
```

But this time, there is no Makefile to read.. and hence, no `app` target to run.

I guess the autopatch feature comes in handy here (haven't looked into it yet).. but this could still be addressed, if nothing else, print a descriptive error message (instead of letting users like myself try to figure out what's wrong :p ).

+1 on your last commit, though ;) ( b23a88ada3ad9d38bdc90c1f080fd1ffe396703b )

kaos:
Print helpful error message in case a dependency doesn't have a Makefile to build it.

### DIFF
--- a/core/deps.mk
+++ b/core/deps.mk
@@ -35,7 +35,7 @@ PKG_FILE_URL ?= https://raw.githubusercontent.com/ninenines/erlang.mk/master/pac
 deps:: $(ALL_DEPS_DIRS)
 	@for dep in $(ALL_DEPS_DIRS) ; do \
 		if [ -f $$dep/GNUmakefile ] || [ -f $$dep/makefile ] || [ -f $$dep/Makefile ] ; then \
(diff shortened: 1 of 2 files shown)


essen:
Merged, thanks!

PR #233, opened 2015-04-29
rename wpool to worker_pool

ingwinlu:


essen:
I should have seen it ahah.

ingwinlu:
I should have seen it first as I was actually using it at the time ;), again sorry!

PR #234, opened 2015-04-29
Packages: rename exsplus to exsplus128

jj1bdx:
The following package is renamed and the packages.* files are duly updated. The renamed package: https://github.com/jj1bdx/exsplus128/


essen:
Merged, thanks!

PR #223, opened 2015-03-28
Elixir plugin

deadtrickster:
Elixir files can be placed to `src` or `lib`. 
Also this part of Elixir quickstart might be useful: http://elixir-lang.org/crash-course.html#adding-elixir-to-existing-erlang-programs

PR #237, opened 2015-05-11
Elixir plugin

deadtrickster:
Elixir files can be placed to src or lib. 
Also this part of Elixir quickstart might be useful: http://elixir-lang.org/crash-course.html#adding-elixir-to-existing-erlang-programs

fix for #223

### DIFF
--- a/build.config
+++ b/build.config
@@ -23,6 +23,7 @@ plugins/ci
 plugins/ct
 plugins/dialyzer
 plugins/edoc
+plugins/elixir
 plugins/elvis
 plugins/erlydtl
 plugins/escript
--- a/packages.v1.tsv
+++ b/packages.v1.tsv
@@ -22,7 +22,7 @@ eganglia	https://github.com/inaka/eganglia	https://github.com/inaka/eganglia	Erl
 ehsa	https://bitbucket.org/a12n/ehsa	https://bitbucket.org/a12n/ehsa	Erlang HTTP server basic and digest authentication modules
 ej	https://github.com/seth/ej	https://github.com/seth/ej	Helper module for working with Erlang terms representing JSON
 eleveldb	https://github.com/basho/eleveldb	https://github.com/basho/eleveldb	Erlang LevelDB API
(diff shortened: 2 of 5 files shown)


essen:
Hey, sorry just thought of this now. We can't do 'lib/' because this conflicts with Erlang's own lib (ERL_LIBS) that I plan to use to enable multi application repositories. Do you think that's problematic if it's not handled?

deadtrickster:
Elixir (well mix) people place their code in lib/ usually. I personally prefer src/ at least for erlang.mk based stuff.

essen:
Yeah but we're not going to compile whole Elixir projects here. Let's start with src/ since that fits your needs and can definitely be useful for the odd module, and then we'll see if people need more.

essen:
Fixed the issue with the cover test suite on master if you want to add a test.

essen:
I think I should also add support for Elixir dependencies.

deadtrickster:
speeking of elixir templates, what you suggestions?

zuiderkwast:
In which way would 'lib/' conflict with ERL_LIBS? ERL_LIBS is a list of directories that contain applications while 'lib/' is a directory inside an application. I can't see how these would conflict.

essen:
I have plans to allow repositories to contain more than one application and ./lib/ would be the default directory (same as OTP).

essen:
For info I can't merge this yet because we are patching rebar out and that breaks the compilation of Elixir. Adding it to #265 for now.

michaelklishin:
@essen this could be quite useful for some future improvements in RabbitMQ, for which Elixir makes more sense. Is this still not ready for merging?

essen:
This probably won't be merged because this exists: https://github.com/botsunit/elixir.mk

Not sure where Elixir would make more sense, so in case it's DSL related take note that this also exists: https://github.com/ninenines/lfe.mk

essen:
Closing in favor of the Elixir plugin. Thanks!

PR #243, opened 2015-05-13
Add support for rebar "raw" deps

zerotao:
rebar deps have support for "raw" packages i.e. packages that are
not intended to be built.  e.g.

{deps, [
  {json_patch_tests, "", {git, "git://github.com/json-patch/json-patch-tests.git"}, [raw]}
]}

This change adds support for those packages
when AUTOPATCHing and as a result also addes support for download only
dependencies at the top level via DEPS_FETCH_ONLY += <package>

### DIFF
--- a/core/deps.mk
+++ b/core/deps.mk
@@ -15,6 +15,7 @@ REBAR_DEPS_DIR = $(DEPS_DIR)
 export REBAR_DEPS_DIR
 
 ALL_DEPS_DIRS = $(addprefix $(DEPS_DIR)/,$(DEPS))
+BUILD_DEPS_DIRS = $(filter-out $(addprefix $(DEPS_DIR)/,$(DEPS_FETCH_ONLY)), $(ALL_DEPS_DIRS))
 
 ifeq ($(filter $(DEPS_DIR),$(subst :, ,$(ERL_LIBS))),)
 ifeq ($(ERL_LIBS),)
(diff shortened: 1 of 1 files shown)


essen:
This will break at least one of the packages in the index I think (but which one, I don't really remember).

essen:
I suppose what we could do *instead* is to detect cases where we clearly don't have an Erlang application, and not do anything when that happens.

What's the application that has this example above?

essen:
By not do anything I mean we create an empty Makefile if none exist to no-op.

zerotao:
I ran into this when integrating https://github.com/talentdeficit/json into a project.  I'm perfectly willing to go with the empty Makefile.

Is the project this would break using 'raw' but expecting it to be built anyway?

essen:
I will have to get back to you on that. I am currently working on improving compatibility and have seen more than a hundred projects today so my memory is a little fuzzy. I will ping the ticket when I know for sure what we need to do.

essen:
I think for now I will detect if there is an src/ folder and if not then put an empty Makefile. How does that sound?

zerotao:
We might also look for an apps folder for the multi app approach but the above works for me.  I would be happy to implement it if you like.

essen:
I don't think apps/ layouts are allowed in dependencies even in rebar. Haven't seen a single one in the near 200 projects that erlang.mk now compiles.

zerotao:
ok, that's reasonable.

essen:
Fixed in c0c2863a21a2eee46c69e5a2ce275c91aa2f2ba3 and the package json added to the index. Thanks!

zerotao:
thanks

PR #244, opened 2015-05-14
Build NIF C code also on tests

jj1bdx:
* This patch ensures building the NIF code under`c_src`
  even on running `make tests`

* When `c_src/Makefile` exists:
    * Split the Makefile invocation dependency into `app-c_src`
    * Invoke `app-c_src` also in `test-build` as well as `app`
* When `c_src/Makefile` does not exist:
    * Add `$(C_SRC_ENV) $(C_SRC_OUTPUT)` dependencies to `test-build`

PR #246, opened 2015-05-15
Fix test/Makefile for FreeBSD shell redirection compatibility

jj1bdx:
* FreeBSD shell does not accept `&>` as the redirect operator;
  Explicitly rewrite `&>/dev/null` to `>/dev/null 2>&1`
  as a workaround
* This will not break other shells (confirmed on OS X 10.10.3)

### DIFF
--- a/test/Makefile
+++ b/test/Makefile
@@ -14,12 +14,12 @@ V ?= 0
 ifeq ($V,0)
 	# Show info messages only
 	t = @
-	v = V=0 &>/dev/null
+	v = V=0 >/dev/null 2>&1
 	i = @echo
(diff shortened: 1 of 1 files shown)


jj1bdx:
See also #244

essen:
Looks fine however it seems I have broken something so I will get back to you.

jj1bdx:
OK. I discovered `&>` is bash dependent. FYI. You can merge this together with other stuff.

essen:
Unrelated to your PR so merging, thanks!

Nothing broken per se either, just messages that shouldn't be there.

jj1bdx:
:+1:

PR #247, opened 2015-05-16
move nodefinder dep to erlang.mk compatible source

ingwinlu:
since the upstream repo did not react to a PR I forked and merged my patch to master there


essen:
Did you try with current erlang.mk? The original works for me.

ingwinlu:
indeed it seems to build fine now, while requiring AUTOPATCH before, good work!

essen:
AUTOPATCH was removed and is applied by default now. Only a couple issues to fix and I can claim it compiles everything (at least everything I can throw at it).

PR #258, opened 2015-05-30
Able to use external configuration file for CT

ethrbh:
hello,

I have made a small improvement in ct.mk. This change pass external configuration files to CT when it starting. The external configuration files have two requirements:
 - the files must be placed in the test/ folder
 - must set .cfg extension for the files

Please merge the PR once you have reviewed and all looks good.

thanks for your help,
/Robi

### DIFF
--- a/erlang.mk
+++ b/erlang.mk
@@ -4,7 +4,8 @@ ERLANG_MK_BUILD_CONFIG ?= build.config
 ERLANG_MK_BUILD_DIR ?= .erlang.mk.build
 
 erlang.mk: bootstrap
-	git clone https://github.com/ninenines/erlang.mk $(ERLANG_MK_BUILD_DIR)
+	#git clone https://github.com/ninenines/erlang.mk $(ERLANG_MK_BUILD_DIR)
+	git clone https://github.com/ethrbh/erlang.mk.git $(ERLANG_MK_BUILD_DIR)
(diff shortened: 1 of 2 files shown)


essen:
What about just doing CT_OPTS = -config yourconfigfile.config ?

ethrbh:
hello,

How do you mean to use this variable? You mean to use in command line when starting ct, or inside the makefile?
I tried to use it like this, but it did not work. I guess I did something wrong, and that is why I made another solution.
    make ct CT_OPTS = -config test/basic_SUITE_1.cfg test/basic_SUITE_2.cfg

thanks for your help,
/Robi

essen:
If you do it in the command line instead of your Makefile (not sure why you wouldn't put it in the Makefile though), you need to quote the value, ie:

make ct CT_OPTS="-config a.cfg b.cfg"

ethrbh:
hello,

Thanks, now it works with the quote. But I preferred to have -config in the Makefile instead. So, do I understood you right, that you tried say in your 1st comment, to use CT_OPTS var instead of introduce a new one, what I did?

thanks,
/Robi

essen:
In your Makefile you can put:

CT_OPTS = -config a.cfg b.cfg

ethrbh:
hello,

Oh my God! Why did not think about of that?! Of course it is working in that way too. In this case I think I can abandon my changes, and I also can close the PR.

thanks for your help,
/Robi

essen:
No worries. :-)

PR #269, opened 2015-06-09
cover.mk: all.coverdata made one-liner

jpuigm:
As per https://github.com/ninenines/erlang.mk/pull/204#discussion_r31495209

### DIFF
--- a/plugins/cover.mk
+++ b/plugins/cover.mk
@@ -78,9 +78,7 @@ coverdata-clean:
 
 # Merge all coverdata files into one.
 all.coverdata: $(COVERDATA)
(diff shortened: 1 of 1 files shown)


essen:
Thanks, EUC travels started so I'll see when I can try. :-)

essen:
OK not exactly what I was thinking of. I've been moving to a variable ending with .erl that has the Erlang code, and then the $(call erlang, ...) to run it. For example: https://github.com/ninenines/erlang.mk/blob/master/plugins/triq.mk

essen:
I've done the change I mentioned and converted the cover tests, fixed a few issues at the same time. I'll consider this resolved. Thanks!

PR #272, opened 2015-06-11
xrefr upgrade

elbrujohalcon:


essen:
Cool, looks good. EUC so merge time is undefined. Any chance I can get an Elvis update too? Seems it's late.

elbrujohalcon:
yeah… as soon as we reach a new release milestone with elvis, you'll get an upgrade as well

essen:
Merged, thanks.

PR #278, opened 2015-06-19
dep_autopatch: Ensure ['s `!=` operator has operands

dumbbell:
Consider the following test:
```sh
elif [ 0 != `find ... | xargs grep -ci rebar` ]; then
```

find(1) may return no file at all and xargs has nothing on its stdin. In this case, GNU xargs (from findutils) still executes the given command and grep(1) returns `0`.

However, FreeBSD's xargs for instance does not run the command at all if there is nothing on stdin. Therefore nothing is printed on stdout and the test becomes:
```sh
elif [ 0 != ]; then
```

This triggers a warning from the shell:
```
[: !=: argument expected
```

Prepending both operands with a literal `x` fixes the problem.

For consistency's sake, I modified the two other tests as well, though I didn't have any problem with them.

### DIFF
--- a/core/deps.mk
+++ b/core/deps.mk
@@ -67,11 +67,11 @@ distclean:: distclean-deps distclean-pkg
 # in practice only Makefile is needed so far.
 define dep_autopatch
 	if [ -f $(DEPS_DIR)/$(1)/Makefile ]; then \
-		if [ 0 != `grep -c "include ../\w*\.mk" $(DEPS_DIR)/$(1)/Makefile` ]; then \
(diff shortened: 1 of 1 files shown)


essen:
Yeah I need to install FreeBSD to make sure these things don't happen. :-)

dumbbell:
That would be a wise decision ;p

essen:
OK I will run a test of all packages with this and next patch. Should take 2 to 3 hours.

essen:
And it's also in, thanks!

essen:
OK this fixes the error but problem is it does not enter the if when it should, I'll figure another way.

dumbbell:
You're right, it should enter the `if` when find(1) does not return any file, sorry...

PR #280, opened 2015-06-19
core/deps.mk: Ensure `$(ERLANG_MK_TMP)` is created

dumbbell:
... before writing to it.

This fixes the following error which is logged with a fresh checkout:
```
/bin/sh: cannot create (...)/.erlang.mk/deps.log: No such file or directory
```


essen:
Ah cool, I'll merge that after testing.

essen:
And it's in, thanks!

PR #284, opened 2015-06-26
Fix deps regexp

nevar:
$$$$ after replace by make is $$ and in bash it is pid of process.


essen:
That could be why it stopped working. :-)

I'll try when I get back, thanks!

essen:
Merged, thanks!

PR #293, opened 2015-07-02
add overwritable WS variable...

ingwinlu:
...to allow users to customize what **W**hite**S**pace they want to use for indentation.

Not sure where to add documentation for this though, recommendations?

Usage:
```
WS="    " make -f erlang.mk bootstrap
```

### DIFF
--- a/core/core.mk
+++ b/core/core.mk
@@ -119,6 +119,7 @@ help::
 
 empty :=
 space := $(empty) $(empty)
+tab := $(empty)	$(empty)
 comma := ,
 
 define newline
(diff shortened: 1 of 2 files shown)


essen:
Instead of replacing everything in the templates, I would rather the template use a tab, and the tab is replaced when we create the file. :-)

Don't worry about the doc, it's being rewritten separately.

essen:
So I pushed something today, hope you can feedback soon. Enjoy!

essen:
So I actually pushed something now. Enjoy! :-)

PR #297, opened 2015-07-05
The '-r' option of 'xargs' is illegal on OSX

manifest:
```
 DEP    cowboy
xargs: illegal option -- r
usage: xargs [-0opt] [-E eofstr] [-I replstr [-R replacements]] [-J replstr]
             [-L number] [-n number [-x]] [-P maxprocs] [-s size]
             [utility [argument ...]]
```

### DIFF
--- a/core/core.mk
+++ b/core/core.mk
@@ -170,6 +170,12 @@ core_lc = $(subst A,a,$(subst B,b,$(subst C,c,$(subst D,d,$(subst E,e,$(subst F,
 # @todo On Windows: $(shell dir /B $(1)); make sure to handle when no file exists.
 core_ls = $(filter-out $(1),$(shell echo $(1)))
 
+ifeq ($(PLATFORM),darwin)
+core_xargs = xargs
+else
+core_xargs = xargs -r
+endif
+
 # Automated update.
 
 ERLANG_MK_BUILD_CONFIG ?= build.config
--- a/core/deps.mk
+++ b/core/deps.mk
@@ -67,7 +67,7 @@ define dep_autopatch
 			$(call dep_autopatch2,$(1)); \
 		elif [ 0 != `grep -ci rebar $(DEPS_DIR)/$(1)/Makefile` ]; then \
 			$(call dep_autopatch2,$(1)); \
-		elif [ -n "`find $(DEPS_DIR)/$(1)/ -type f -name \*.mk -not -name erlang.mk | xargs -r grep -i rebar`" ]; then \
+		elif [ -n "`find $(DEPS_DIR)/$(1)/ -type f -name \*.mk -not -name erlang.mk | $(call core_xargs) grep -i rebar`" ]; then \
 			$(call dep_autopatch2,$(1)); \
 		else \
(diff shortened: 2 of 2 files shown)


essen:
Thanks, I'll try it in a few days, I can't access my OSX test VM at the moment. :-)

danielwhite:
Can confirm this works for me.

Funnily enough, when building cowboy, this ends up resulting in an error that looks a lot like ninenines/cowboy#842.

danielwhite:
As an aside, I think this might be unnecessarily complicated. GNU xargs works perfectly fine without the `-r` option/

Commit 75453090 introduced the flag, and mentioned FreeBSD. Is this flag actually needed for FreeBSD support?

manifest:
According to the man page, the -r option does nothing in the FreeBSD version of xargs.

```
     -r	     Compatibility with	GNU xargs.  The	GNU version of xargs runs the
	     utility argument at least once, even if xargs input is empty, and
	     it	supports a -r option to	inhibit	this behavior.	The FreeBSD
	     version of	xargs does not run the utility argument	on empty
	     input, but	it supports the	-r option for command-line compatibil-
	     ity with GNU xargs, but the -r option does	nothing	in the FreeBSD
	     version of	xargs.
```

essen:
See this commit instead: 0040f93ccd5f4aa6db719741fcd6335c48d1390d

essen:
The -r option was added so that it works on both, but I guess OSX is even more different. :-)

manifest:
I don't see any description, even mention of the `-r`  option of xargs utility on OSX. But it looks like its BSD version works the same way as GNU xargs works with `-r` options. If the standard input doesn't contain any nonblanks, the command won't be run. As for OSX, they doesn't just ignore the `-r` option as FreeBSD does, they drop it completely out.

I run a simple test which, I think, confirms that.
```sh
## Linux

$ ls -1
1

$ find . -name 1 | xargs echo '--xargs utility has been run--';
--xargs utility has been run-- ./1

$ find . -name 2 | xargs echo '--xargs utility has been run--';
--xargs utility has been run--

$ find . -name 2 | xargs -r echo '--xargs utility has been run--';
$



## OSX

$ ls -1
1

$ find . -name 1 | xargs echo '--xargs utility has been run--';
--xargs utility has been run-- ./1

$ find . -name 2 | xargs echo '--xargs utility has been run--';
$

$ find . -name 2 | xargs -r echo '--xargs utility has been run--';
xargs: illegal option -- r
usage: xargs [-0opt] [-E eofstr] [-I replstr [-R replacements]] [-J replstr]
             [-L number] [-n number [-x]] [-P maxprocs] [-s size]
             [utility [argument ...]]
$
```

essen:
Well, considering xargs is called exactly one time in the whole file, and that find has an --exec option, perhaps we should just drop xargs and use --exec as it looks like a better solution for portability.

manifest:
I think, it's a good idea. :)

manifest:
Made another pull request replacing `xargs` by `find -exec`.

essen:
Closing this considering the other PR looks good. Still no SSH access though so not merging just yet. Just bored, closing tickets. :-)

PR #301, opened 2015-07-12
Replace 'xargs' with 'find -exec' for portability reasons

manifest:
As result of this conversation on same issue:
https://github.com/ninenines/erlang.mk/pull/297


essen:
Looks good, thanks! Can't merge now due to no SSH access, have to wait next week.

essen:
Merged, thanks!

PR #320, opened 2015-08-08
symbolic links to deps

brigadier:
Allows to make sybolic links to deps. Same as `cp` option, but `ln -s -r` instead of `cp -R`.


brigadier:
closing due to commited compiled erlang.mk. Will open a new request in a minute

PR #321, opened 2015-08-08
allow creating symbolic links to local deps

brigadier:
Allows to make symbolic links to deps. Same as `cp` option, but `ln -s -r` instead of `cp -R`.

### DIFF
--- a/core/deps.mk
+++ b/core/deps.mk
@@ -495,6 +495,8 @@ define dep_fetch
 		svn checkout -q $(3) $(DEPS_DIR)/$(1); \
 	elif [ "$(2)" = "cp" ]; then \
 		cp -R $(3) $(DEPS_DIR)/$(1); \
+	elif [ "$(2)" = "ln" ]; then \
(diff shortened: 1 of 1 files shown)


essen:
Why -r though? That will make it break if you move the folder, no?

brigadier:
Yes, this will make it break when folder gets moved.
I have no any special preferences neither for -r nor against it

essen:
I rather not have it break. :-)

brigadier:
k, removed -r flag

essen:
OK I was experimenting with this and I think it's best NOT to add this. It adds a lot of complexity and is not adequate IMO since autopatch will overwrite a number of files in your local project. I don't think we ever want this behavior.

This doesn't mean symbolic links won't work; you can create them, even automatically if you need to. Just that they should be handled in a different manner from the normal dep fetching. I will probably write a tutorial at some point about using symbolic links during development.

In the meantime, closing this. Thanks!

PR #323, opened 2015-08-15
Added EUNIT_FLAGS to add flexibility to eunit

cecalab:
I added EUNIT_FLAGS so that command-line arguments can be supplied to eunit. For example, to load the application's vm.args and sys.config, add this line to your makefile:

EUNIT_FLAGS += -args_file 'rel/vm.args' -config 'rel/sys.config'

### DIFF
--- a/plugins/eunit.mk
+++ b/plugins/eunit.mk
@@ -6,6 +6,7 @@
 
 # Configuration
 
+EUNIT_ERL_OPTS ?=
 EUNIT_OPTS ?=
 
 # Core targets.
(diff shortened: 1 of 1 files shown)


essen:
Thanks. The erlang.mk file should not be submitted, it's the other files that should (plugins/eunit.mk in your case). Hard to know what changed right now. :-)

cecalab:
I committed the plugin, but the test seems to be failing when trying to install autoconf:
sudo apt-get install autoconf2.59

essen:
Rebase to master I fixed it there. It will still fail, but after 1 hour or so of testing. :-)

cecalab:
Yep, failed after an hour following rebase.

essen:
No worries, you didn't break anything. :-)

essen:
I'm not sure about the variable name, need to think about it. We use "OPTS" everywhere so it's a bit weird to start using "FLAGS".

Also please don't commit the erlang.mk file. The one in the repo is just a bootstrap file. :-)

cecalab:
How about EUNIT_ERL_OPTS? Ok, I'll check out the master erlang.mk and repair.

essen:
Yep EUNIT_ERL_OPTS might be good.

cecalab:
Ok, I committed EUNIT_ERL_OPTS and reverted erlang.mk

essen:
Looks good, so, whenever I'm ready now. :-)

essen:
Pushed the equivalent of this patch against today's master. Thanks!

PR #327, opened 2015-08-18
Fix build config renaming

zerotao:
Currently build.config can be placed in a different directory via
something like:

  ERLANG_MK_BUILD_CONFIG = tools/build.config

but if the name is changed:

  ERLANG_MK_BUILD_CONFIG = .erlang-mk.config

it will be ignored.


essen:
Good catch. :-)

essen:
Merged, thanks!

PR #330, opened 2015-08-25
Support $(EXTRA_SOURCES) to list generated sources

dumbbell:
Files listed in this variable can be Erlang modules, headers, MIB, ASN.1 files or `xrl`/`yrl` sources. Their corresponding target will be evaluated before the compilation starts.

They are removed during `make clean-app`.

### DIFF
--- a/core/erlc.mk
+++ b/core/erlc.mk
@@ -70,7 +70,7 @@ define app_file
 endef
 endif
 
-app-build: erlc-include ebin/$(PROJECT).app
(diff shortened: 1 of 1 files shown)


essen:
OK I'll think how that will fit with the other changes I mentioned.

Also the tests will fail, don't worry about it.

essen:
Shouldn't be needed anymore, so closing, thanks!

PR #331, opened 2015-08-25
New variable ERLANG_MK_DISABLE_PLUGINS to disable plugins

dumbbell:
For instance, one can disable the `ct` and `edoc` plugins by setting:
```
ERLANG_MK_DISABLE_PLUGINS = ct edoc
```

### DIFF
--- a/plugins/asciidoc.mk
+++ b/plugins/asciidoc.mk
@@ -1,6 +1,8 @@
 # Copyright (c) 2015, Loïc Hoguin <essen@ninenines.eu>
 # This file is part of erlang.mk and subject to the terms of the ISC License.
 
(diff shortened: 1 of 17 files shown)


dumbbell:
`build.config`, though less handy, does the job.

essen:
Hey, I still want something like this, so I'll leave it open for now. I'm just not sure it should take this form yet.

essen:
Note that while there's still nothing for this specifically, now edoc is only enabled on "make docs" by default if a "doc/overview.edoc" file exists. This might solve your own issue.

I am still interested in disabling entirely so I will keep this open a little more.

essen:
Closed in favor of #543. Thanks!

PR #334, opened 2015-08-27
.gitignore: Ignore Vim .swp files

dumbbell:


essen:
Wow, sorry, I just removed the file. Probably won't merge cleanly anymore. :-)

dumbbell:
I updated the pull request with only the change to `.gitignore`.

essen:
As for adding these to gitignore, I actually like to see in "git status" them because then I can either remove in one go (not have to deal with vim) or find what I might have lost if the computer shut down previously. So I won't be merging. :-) Sorry for the waste of time.

dumbbell:
That's ok, I understand :)

PR #340, opened 2015-09-02
app.asciidoc: Document how to handle generated source files

dumbbell:


doc/src/guide/app.asciidoc
```diff
@@ -225,7 +225,7 @@ NOTE: This section is about the dependency tracking between files
 inside your project, not application dependencies.
 
 Erlang.mk keeps track of the dependencies between the different
-files in your project. This information is kept in the '$(PROJECT).d'
+files in your project. This information is kept in the `$(PROJECT).d`
```
essen: So this change I will revert because I am referring to the 'file' rather than the `variable`. It doesn't show on github properly because github has a different tooling, but there's a clear difference in the generated PDF.

essen:
Merged, thanks!

PR #341, opened 2015-09-02
Double colon for source files

dumbbell:


essen:
Incidentally, this leftover touch was also making things twice slower. Pretty cool!

essen:
OK, thought so, there's still some issues.

Try touching an include file, and then run 'make' again. It won't be detected properly, despite the dep file having the rule.

That's probably why the touch erl files was there.

dumbbell:
I believe the commit I just added address the "try touching an include file" problem.

Now, `$(PROJECT).d` only lists files having dependencies. The rules run `touch` on them.

essen:
Pushed a few more changes, including merging of this (before the extra commit, since I had that already, though in my case it was making an infinite loop :-) until I remove ::).

I need to do further testing but I think this is looking good.

Thanks!

PR #343, opened 2015-09-03
Generate `$(PROJECT).d` before calling `make app-build`

dumbbell:
This ensures all generated dependencies are prepared and the `.d` file is ready before the actual build is started.


essen:
Merged, thanks!

PR #345, opened 2015-09-04
deps.mk: Do not remove $(DEPS_DIR) if $(SKIP_DEPS) is set

dumbbell:
I suppose `$(SKIP_DEPS)` is here to let people download dependencies themselves. Therefore, `$(DEPS_DIR)` should not be removed when `make distclean` is called.

### DIFF
--- a/core/deps.mk
+++ b/core/deps.mk
@@ -571,8 +571,12 @@ endef
 
 $(foreach dep,$(BUILD_DEPS) $(DEPS),$(eval $(call dep_target,$(dep))))
 
+ifneq ($(SKIP_DEPS),)
+distclean-deps: ; @echo -n
+else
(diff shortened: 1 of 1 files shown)


dumbbell:
In the context of RabbitMQ, as we work on many repositories related to each other, we will probably use `$(SKIP_DEPS)` so we can clone all repositories in one place (what the current `rabbitmq-public-umbrella` does).

But if `make distclean` removes this directory, it will make this impossible.

essen:
Makes perfect sense. Can it wait a few days? I'd like to merge it at the same time I write the deps docs and their related tests.

dumbbell:
Sure!

essen:
Pushed something similar but with a test. Closing, thanks!

PR #351, opened 2015-09-07
deps.mk: Support multiple commits, used as fallbacks

dumbbell:
Here is an example:
```
dep_mydep = git https://... stable-branch master
```

In this case, erlang.mk will clone `mydep` and checkout the `stable-branch` branch. If this reference does not exist, it tries `master`.

Of course, if none of the references are valid, it aborts.

### DIFF
--- a/core/deps.mk
+++ b/core/deps.mk
@@ -475,12 +475,18 @@ endef
 
 define dep_fetch_git
 	git clone -q -n -- $(call dep_repo,$(1)) $(DEPS_DIR)/$(call dep_name,$(1)); \
-	cd $(DEPS_DIR)/$(call dep_name,$(1)) && git checkout -q $(call dep_commit,$(1));
+	cd $(DEPS_DIR)/$(call dep_name,$(1)) && ( \
+	$(foreach ref,$(call dep_commits,$(1)), \
+	  git checkout -q $(ref) >/dev/null 2>&1 || \
+	  ) (echo "error: no valid pathspec among: $(call dep_commits,$(1))" 1>&2 && false) )
 endef
 
 define dep_fetch_hg
 	hg clone -q -U $(call dep_repo,$(1)) $(DEPS_DIR)/$(call dep_name,$(1)); \
-	cd $(DEPS_DIR)/$(call dep_name,$(1)) && hg update -q $(call dep_commit,$(1));
+	cd $(DEPS_DIR)/$(call dep_name,$(1)) && ( \
+	$(foreach ref,$(call dep_commits,$(1)), \
+	  hg update -q $(ref) >/dev/null 2>&1 || \
+	  ) (echo "error: no valid pathspec among: $(call dep_commits,$(1))" 1>&2 && false) )
 endef
 
 define dep_fetch_svn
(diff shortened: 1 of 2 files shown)


essen:
Thanks.

If anyone is interested by this kind of feature, please raise your voice. :-)

dumbbell:
Do people usually visit issues/PR and give their opinion? :)

Let me try to convince you!

This feature is quite specific to users working on multiple related projects at the same time (of course, this is the case for us with RabbitMQ). This is of no use for users working on a single project having third-party dependencies.

Here is our workflow: RabbitMQ is split into multiple repositories: one for the broker itself, one for each plugins, one for common Erlang modules (shared with the broker, the plugins and the Erlang client), and so on. We maintain two branches for all the projects composing RabbitMQ: `master` which is the development version (the future X.(Y+1).0) and `stable` which is the branch we tag for each release (X.Y.(Z+1)). Bug fixes go to `stable`, new features to `master`. `stable` is merged to `master` after each bug fix, so they are part of the future major release. `master` is merged to `stable` when we are ready to produce the new X.(Y+1).0.

Sometimes, we need to prepare a patch based on an older release for paying customers. In this case, we need to checkout a particular tag, eg. `rabbitmq_v3_5_1`, in all components.

Currently, when we want to checkout eg. `stable` for all components, we use the rabbitmq-public-umbrella project (the same "wrapper" used to do the initial git clone of all components), because our homebrew Makefiles do not have proper dependency handling like erlang.mk.

This looks like a common workflow to me when you have a piece of software split in several components/repositories.

With the adoption of erlang.mk, this umbrella thing looses interest: if I clone the shovel plugin, I expected `make` to download the relevant dependencies of course (the common Erlang modules, the Erlang client, and, for the tests, the broker). One way to do this with the current erlang.mk is to put the exact branch, tag or commit in the top-level Makefile. However, this means we will have to constantly update all Makefiles (before each tag) and/or resolve conflicts (when merging `master` and `stable`).

The proposed patch allows to put this logic in the Makefile itself, instead of relying on manual error-prone steps. In my test, this looks like this:

```make
# ...

current_rmq_ref := $(shell git symbolic-ref -q --short HEAD || git describe --tags --exact-match)
base_rmq_ref := $(shell git merge-base --is-ancestor $$(git merge-base master HEAD) stable && echo stable || echo master)

dep_rabbit_common = git https://.../rabbitmq-common.git $(current_rmq_ref) $(base_rmq_ref)

include erlang.mk
```

I feel it can be useful to other larger projects. Furthermore, the patch is small and doesn't break compatibility.

essen:
Yeah features that many people need tend to have many people posting about it.

The problem with the patch isn't that it's big or backward incompatible, it's that it's git-only and possibly rabbitmq-only.

For the time being, I'll go with a patch that allows overriding those functions (probably just ifdef around) rather than changing them directly. Then if time allows perhaps we can do this for all relevant fetch methods, which will make more sense to include.

That said, perhaps something more useful to everyone would be to not just look at different tags, but to allow listing more than one "source", ie not just the commit but the repo also. This would allow providing mirrors of repositories, the mirrors not necessarily using the same fetch method (for example one could be a local folder, then a remote archive, then a git repo). And *that* I am sure will be useful to a lot of people, while still fitting your use case.

essen:
I'll probably take a look at adding these mirror things when I get to document deps. This sounds quite useful.

essen:
Shouldn't be incompatible with your use, we just have to not fetch the repo again if it didn't change.

essen:
Using this ticket to put thoughts about related things.

Currently we have DEPS=cowboy and then dep_cowboy="git repo commit".

One thing I want to add is the ability to override the commit number:

```
DEPS = cowboy
dep_cowboy = git repo abcdef
dep_cowboy_commit = fedcba
```

In this case this would use the commit fedcba instead of the one specified normally. Main reason for wanting this is so you can just say DEPS=cowboy and then override the commit, and it will take all info from the packages except the commit number. Another possible use is adding a "lock file" that would keep track of the actual commit number used by the dependencies (that you could commit to have reproducible builds). The benefits are obvious.

essen:
ON TOPIC: why not just define a special fetch method, say git-whatever, which does your functions? That sounds much cleaner and easier to support.

dumbbell:
I finally added a small Makefile doing that to all projects.

PR #353, opened 2015-09-08
erlc.mk: Do not try to build .d or .app files...

dumbbell:
... if there is no Erlang application or library.

### DIFF
--- a/core/erlc.mk
+++ b/core/erlc.mk
@@ -40,6 +40,8 @@ asn1_verbose = $(asn1_verbose_$(V))
 mib_verbose_0 = @echo " MIB   " $(filter %.bin %.mib,$(?F));
 mib_verbose = $(mib_verbose_$(V))
 
+ifneq ($(wildcard src/),)
+
 # Targets.
 
 ifeq ($(wildcard ebin/test),)
@@ -79,7 +81,6 @@ app-build: ebin/$(PROJECT).app ; @echo -n
 
 # Source files.
 
-ifneq ($(wildcard src/),)
 ERL_FILES = $(sort $(call core_find,src/,*.erl))
 CORE_FILES = $(sort $(call core_find,src/,*.core))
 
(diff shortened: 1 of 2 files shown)


essen:
I prefer a shorter commit message and not defining new variables (many wildcards is OK, it'll hit the Make file cache anyway).

dumbbell:
Ok, let me change that.

core/erlc.mk
```diff
@@ -43,10 +43,18 @@ mib_verbose = $(mib_verbose_$(V))
 # Targets.
 
 ifeq ($(wildcard ebin/test),)
```
essen: Honestly, reading the resulting file, it sounds like everything from above this line to the end of the file (including clean, since it's after all not an Erlang app) should be in a single $(wildcard src/).

With your changes, all that's left is the app_file definition and the empty app-build, the first being used only in this file and the second would not be necessary anymore if everything is inside a single if block.

test/Makefile
```diff
@@ -77,7 +77,7 @@ clean-core: clean-core-app clean-core-upgrade
 
 # Core: Building applications.
 
-CORE_APP_CASES = asn1 error generate-erl generate-erl-include generate-erl-prepend hrl hrl-recursive mib no-makedep xrl xrl-include yrl yrl-include
+CORE_APP_CASES = no-app asn1 error generate-erl generate-erl-include generate-erl-prepend hrl hrl-recursive mib no-makedep xrl xrl-include yrl yrl-include
```
essen: Alphabetical order. :-)

test/Makefile
```diff
@@ -85,6 +85,17 @@ CORE_APP_CLEAN_TARGETS = $(addprefix clean-,$(CORE_APP_TARGETS))
 
 core-app: $(CORE_APP_TARGETS)
 
+core-app-no-app: build clean-core-app-no-app
```
essen: Same here. :-)

dumbbell:
The new test is sorted now :)

dumbbell:
Patch updated, I think I addressed all your comments.

essen:
Merged, thanks!

PR #354, opened 2015-09-08
core.mk: Allow to take erlang.mk from an alternate Git repository

dumbbell:
The goal is to be able to run `make erlang-mk` and fetch `erlang.mk` from a fork on GitHub for instance, while waiting for patches to be reviewed.

### DIFF
--- a/core/core.mk
+++ b/core/core.mk
@@ -173,13 +173,15 @@ core_ls = $(filter-out $(1),$(shell echo $(1)))
 
 # Automated update.
 
+ERLANG_MK_REPO ?= https://github.com/ninenines/erlang.mk
+ERLANG_MK_COMMIT ?=
 ERLANG_MK_BUILD_CONFIG ?= build.config
 ERLANG_MK_BUILD_DIR ?= .erlang.mk.build
 
 erlang-mk:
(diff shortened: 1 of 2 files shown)


essen:
OK but let's keep the same names across the whole project if possible: ERLANG_MK_REPO and ERLANG_MK_COMMIT would do.

dumbbell:
Thanks, I updated the patch.

essen:
Thanks. We also need a test for that. I'll think about how to test it.

dumbbell:
I added a test case.

essen:
There's a special category of tests for erlang.mk upgrades later in the file. core-app-* is just about compiling/cleaning. :-)

dumbbell:
Right, fixed :)

essen:
Looks good, will try.

essen:
Merged, thanks!

essen:
Should have looked before merging. CI failed. :-)

essen:
Fixed.

dumbbell:
Thank you. Sorry, I did not see the problem as I have a "normal" checkout, as opposed to what circleci does.

PR #355, opened 2015-09-09
test/Makefile: Use $(MAKE) instead of hardcoding `make`

dumbbell:


essen:
That's what I get for writing tests when I'm sleepy.

essen:
Merged, thanks!

PR #362, opened 2015-09-14
core/deps.mk: New targets `fetch-deps` and `list-deps` to fetch/list deps recurvively

dumbbell:
Therefore, it lists direct dependencies, dependencies' dependencies, and so on.

### DIFF
--- a/build.config
+++ b/build.config
@@ -37,3 +37,6 @@ plugins/xref
 
 # Plugins enhancing the functionality of other plugins.
 plugins/cover
+
+# Core modules which can use variables from plugins.
+core/deps-tools
--- a/core/core.mk
+++ b/core/core.mk
@@ -121,6 +121,8 @@ help::
 		"  all           Run deps, app and rel targets in that order" \
 		"  app           Compile the project" \
 		"  deps          Fetch dependencies (if needed) and compile them" \
+		"  fetch-deps    Fetch dependencies recursively (if needed) without compiling them" \
+		"  list-deps     List dependencies recursively on stdout" \
 		"  search q=...  Search for a package in the built-in index" \
 		"  rel           Build a release for this project, if applicable" \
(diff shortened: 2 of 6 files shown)


dumbbell:
This push request is ready for review. Sorry, I changed it several times after the initial push.

ash-lshift:
This doesn't work for deps which use rebar because it is `test -f`-ing for `erlang.mk`. `erlang.mk` patches their Makefile but doesn't create an `erlang.mk` alongside it.

dumbbell:
@ash-lshift: Thanks for the report, I changed the patch to check for the inclusion of `erlang.mk` instead.

essen:
Something must be wrong. When I test this on the RabbitMQ umbrella, running "make list-deps" takes about a minute.

dumbbell:
I believe the performance issue is now fixed. Also, the code is much simpler.

The problem was that `list-deps` called `fetch-deps` recursively, the latter being also recursive. This led to way too much repository traversals for doing the same thing all over again. Now, `fetch-deps` fills a single output file which is just printed to stdout by `list-deps`.

The patch is ready for another review.

essen:
Running CI now.

essen:
Tests still good.

essen:
I will try it manually on a couple projects but otherwise it should be good.

dumbbell:
Cool, thanks! If you are ok with the behavior and naming, I will write some documentation.

essen:
I'm fine with it. In the future we'll probably want a "make show-deps" to have some details about each dep, but that functionality would have to be built on top of this patch.

Waiting for docs then!

dumbbell:
Do you mean `show-deps` would print more detailed informations, compared to the simple list printed by `list-deps`?

dumbbell:
I added new section to `deps.asciidoc`.

essen:
Merged, thanks!

PR #368, opened 2015-09-21
[READY] prettify generated .app files

ingwinlu:
the new method of generating .app files strips newlines, resulting in very ugly files that still contain loads of whitespace.

- [x] ~~print newlines~~
- [x] ~~remove empty line from id~~

### DIFF
--- a/core/erlc.mk
+++ b/core/erlc.mk
@@ -56,8 +56,8 @@ ifeq ($(wildcard src/$(PROJECT)_app.erl),)
 define app_file
 {application, $(PROJECT), [
 	{description, "$(PROJECT_DESCRIPTION)"},
-	{vsn, "$(PROJECT_VERSION)"},
-	$(if $(IS_DEP),{id$(comma)$(space)"$(1)"}$(comma))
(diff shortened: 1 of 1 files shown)


essen:
Using echo like this won't work everywhere. See circleci failing for example. You have to do the same thing as bootstrap templates do (creating an environment variable and then using it in the echo command).

ingwinlu:
I updated the pr to do the 'export to shell and let it expand' trick that is used in bootstrap to keep the newlines, but this now makes issues as the $(call instructions seem to get ignored.

modules and applications lists are not populated. do I need to 'extract' those calls out of the function?

essen:
Look at how supervisor template is handled when bootstrapping for example. It has a variable. Unfortunately this means moving all the stuff out of the template and defining a lot of temporary variables. Perhaps more crap than I would like to deal with.

Considering your use case was looking into the file to check if everything is OK, perhaps we can include this check as part of the future sanity-check target instead.

essen:
Pushed, thanks!

Issue #242, opened 2015-05-13
Add kjell support for make shell


essen:
Sounds potentially interesting.

essen:
No real need to do this anymore. It's simple enough.

PR #382, opened 2015-10-04
Add kjell support for make shell

bullno1:
This implements #242.

SHELL_KJELL=1 needs to be specified either through env var, Makefile
or argument.

kjell has some issues with sasl: https://github.com/karlll/kjell/issues/22. It's still not clear how to fix it in kjell but from the thread you can see that there's a workaround by adding `-nouser`. Should I append `-nouser` to `SHELL_OPTS` if `SHELL_KJELL` is enabled? Or provide another variable for user to control this fix? i.e: `SHELL_KJELL_NOUSER = 1` by default.

### DIFF
--- a/plugins/shell.mk
+++ b/plugins/shell.mk
@@ -12,6 +12,7 @@ SHELL_KJELL ?=
 ifeq ($(SHELL_KJELL),1)
 SHELL_DEPS += kjell
 SHELL_ERL = $(DEPS_DIR)/kjell/bin/kjell
+SHELL_OPTS += -nouser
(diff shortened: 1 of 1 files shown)


essen:
If everything else works with -nouser, then by all means append it automatically.

essen:
Looks like -nouser got added upstream.

essen:
I will add SHELL_ERL and make the rest an example in the guide. This should be good enough for now.

essen:
Done. Closing, thanks!

PR #386, opened 2015-10-07
Make sure the ebin folder is in the search path

dparnell:
Hi,

  I'm using a custom tag library in my app and found that erlang.mk wasn't handling this properly.
This patch makes sure that the ebin folder is added to the search path so that erlydtl can load things compiled by the app.

Daniel

### DIFF
--- a/plugins/erlydtl.mk
+++ b/plugins/erlydtl.mk
@@ -16,6 +16,7 @@ dtl_verbose = $(dtl_verbose_$(V))
 
 define erlydtl_compile.erl
 	[begin
+		code:add_patha("ebin"),
 		Module0 = case $(DTL_FULL_PATH) of
 			0 ->
 				filename:basename(F, ".dtl");


essen:
Hmm that's weird because there's a "-pa ebin" for this. I'll get to this issue when I document/add tests for erlydtl. Thanks!

dparnell:
Yeah, I thought it was strange as well. I'm pretty sure it was working in a prior version, but something has changed.

essen:
Yep, it just warrants further investigation. :-)

essen:
This should be fixed, although there's no test for this yet, so not closing the ticket yet.

dparnell:
I tried running "make erlang.mk" to get the latest and unfortunately my custom dtl tag no longer works :( I reverted back to my patched version and the code came back

essen:
Adding a test to the test suite would go a long way to getting this fixed. :-)

essen:
I have added a test with a custom tag, it works as intended. I will push it in a few minutes. Closing this then, thanks!

PR #388, opened 2015-10-07
autopatch: do not run autoreconf if configure script already exists

jeanparpaillon:
Due to autoconf/automake versioning, it may be pertinent to push configure script in a git repository.

If configure script exists in a git repository, then it should not be rebuilt automatically.


essen:
Will need to test this against all existing packages, I'm about to setup daily builds of them again, I'll get back to you.

essen:
Is this not needed anymore?

jeanparpaillon:
I have put this patch in a dedicated branch.

PR #393, opened 2015-10-12
Use '+%F %T%z' to format the timestamp

dumbbell:
Option `--rfc-3339` is specific to GNU Coreutils' date(1). This change fixes the use of `make summary` on other platforms such as FreeBSD.


dumbbell:
To be exact, I still have another problem with `make summary` where the `test/packages` directory is missing, but it's unrelated to this date(1) option.

essen:
Fair enough. Might be worth putting the underscore directly instead of using subst then, if possible.

essen:
Merged, thanks!

PR #394, opened 2015-10-12
Do not try to execute an empty shell script in core-plugins-one test

dumbbell:
The previous code did:
```sh
if `<code that prints nothing on stdout>`; then ...
```

So the test failed but not for the good reason. The forked shell printed:
```
/bin/sh: gmake[2]:: not found
```

Now, both `"Run 'make ...'"` tests are simplified and do not rely on backticks and test(1) anymore. Either the output is inspected or the exit code is checked.

### DIFF
--- a/test/Makefile
+++ b/test/Makefile
@@ -348,7 +348,7 @@ ct: app1
 		"all() -> [testcase1]." \
 		"testcase1(_) -> 42 = m:succ(1)." \
 	 > app1/test/failing_SUITE.erl
-	$t if $(MAKE) -C app1 ct-failing $v ; then false ; fi
+	$t ! $(MAKE) -C app1 ct-failing $v
 	$i "Checking that '$(MAKE) distclean-ct' deletes logs."
 	$t $(MAKE) -C app1 distclean-ct $v
 	$t [ ! -e app1/logs ]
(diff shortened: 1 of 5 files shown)


essen:
Can you confirm that the test failed for you and it was not just the output saying this? Because it should just be the output.

dumbbell:
Yeah, I misinterpreted the message: `$(MAKE) -C $(APP) plugin2` still printed `gmake[2]: Entering directory ...` and the shell tried to execute that. I was tricked by the double colon, making me think the command should be printed between the two colons.

But the fix is still correct: the previous code tries to execute something (the output from `$(MAKE) -C $(APP) plugin2`) which is not a shell script at all.

essen:
Ah I understand the issue, OK thanks.

essen:
I find interesting the use of just "!" (would never have guessed that works) so I will do that change everything at the same time.

dumbbell:
I changed all `if $(MAKE) then false` by `! $(MAKE)`. However, the `core-app-pt-erlc-opts` test fails for me on `master` so I can't be 100% sure of my patch; I'm looking at this.

essen:
Merged, thanks!

PR #395, opened 2015-10-13
erlc.mk: Handle .app generation outside of .erl compilation

dumbbell:
Previously, if no source file were changed, the `ebin/$(PROJECT).app` rule was not evaluated again. Therefore, the `.app` file was not updated, even though, `$(GITDESCRIBE)` or `.app.src` was changed.

Now, the .app file is handled separately and always evaluated. Before overwriting the existing `.app` file though, we check if the content of the new file is different.

In the testsuite, the `.app` file is now unmodified.

### DIFF
--- a/core/erlc.mk
+++ b/core/erlc.mk
@@ -197,12 +197,14 @@ endef
 
 ebin/$(PROJECT).app:: $(ERL_FILES) $(CORE_FILES)
 	$(if $(strip $?),$(call compile_erl,$?))
+
+ebin/$(PROJECT).app::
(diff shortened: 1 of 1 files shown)


essen:
Not generating the app file after you change Erlang files will make those Erlang files recompile every time until a new app file is generated, I believe.

dumbbell:
You're right, I will restore the previous behaviour of always write the `.app` file.

dumbbell:
Oh, I see now that you pushed a different fix. Thank you!

PR #397, opened 2015-10-16
Update bootstrap.mk

Tombert:
Add the `init/2` function so that `erlang.mk` works with modern Cowboy.


essen:
The plan is to use external plugins in Cowboy 2. There's a ticket for that there, and it will be done soon. Thanks!

PR #406, opened 2015-10-22
Fix core-app-pt-erlc-opts so the dependency is built

dumbbell:
The generated Makefile said:
```
PROJECT = core_app_pt_erlc_opts
BUILD_DEPS = my_pt_dep
...
```

However, in `core/deps.mk`, the variable `$(ALL_DEPS_DIRS)` calls `$(dep_name)` to get the name of the dependency, and `$(dep_name)` checks the value of either `$(dep_something)` or
`$(pkg_something_name)`. In this test case, none was set, leading to `$(ALL_DEPS_DIRS)` being empty and no dependency was built.

Adding `dep_my_pt_dep = null` to the generated Makefile is enough, though the fetch method is obviously incorrect. The generated Makefile now looks like:
```make
PROJECT = core_app_pt_erlc_opts
BUILD_DEPS = my_pt_dep
dep_my_pt_dep = null
...
```

### DIFF
--- a/core/deps.mk
+++ b/core/deps.mk
@@ -21,7 +21,7 @@ export DEPS_DIR
 REBAR_DEPS_DIR = $(DEPS_DIR)
 export REBAR_DEPS_DIR
 
-dep_name = $(if $(dep_$(1)),$(1),$(pkg_$(1)_name))
+dep_name = $(if $(dep_$(1)),$(1),$(if $(pkg_$(1)_name),$(pkg_$(1)_name),$(1)))
 dep_repo = $(patsubst git://github.com/%,https://github.com/%, \
 	$(if $(dep_$(1)),$(word 2,$(dep_$(1))),$(pkg_$(1)_repo)))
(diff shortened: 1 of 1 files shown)


dumbbell:
Hmm, `core-deps-fetch-fail-unknown` fails for me too, I didn't spotted it earlier, and circleci confirms it.

This looks like a regression caused by bfa0b2edc142cf32a74bcbb36632052a524c3437. Should `$(dep_name)` return its argument as-is or maybe throw an error?

essen:
I'll take a look. Looks like I botched that commit a bit. :-)

dumbbell:
If `$(dep_name)` returns its argument as a last resort, both tests work:
```
dep_name = $(if $(dep_$(1)),$(1),$(if $(pkg_$(1)_name),$(pkg_$(1)_name),$(1)))
```

dumbbell:
I believe this behaviour of `$(dep_name)` is sensible (compared to throwing an error). I can udpate this pull request if you agree.

essen:
Sounds good.

essen:
Merged, thanks!

PR #408, opened 2015-10-22
Use `:` as a noop instead of `echo -n`

dumbbell:
The `-n` flag is not standard and not portable. It is supported by many Bourne shell to suppress the trailing newline character. But some old shells do not support this and "-n" is printed.

This is the case for instance on some old versions of Mac OS X where `/bin/sh` is Bash 3.2.

### DIFF
--- a/core/core.mk
+++ b/core/core.mk
@@ -82,7 +82,7 @@ all:: deps app rel
 
 # Noop to avoid a Make warning when there's nothing to do.
 rel::
-	$(verbose) echo -n
+	$(verbose) :
 
 check:: clean app tests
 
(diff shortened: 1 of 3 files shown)


dumbbell:
@essen, I know you told me you wanted to use `echo -n` before you added support for Windows/MSYS2 in case `:` was not supported there. I tested `:` with MSYS2 installed yesterday and it worked as expected.

essen:
Fair enough. Do we really need to create a variable for that though?

dumbbell:
I don't think so, but that was to start the discussion, in case you knew another case where `:` was not supported.

essen:
Should work fine in all supported platforms, I believe.

dumbbell:
Ok, patch updated!

essen:
Running tests everywhere to make sure but looks like it's working fine so far.

essen:
Merged, thanks!

Issue #409, opened 2015-10-23
Cover can fail with 'badarith' error


hdima:
In my project I have a module which doesn't contain any functions but only defines a behaviour so cover fails with `badarith` error at https://github.com/ninenines/erlang.mk/blob/master/plugins/cover.mk#L106 because `Y` and `N` are both 0 for the module and it tries to do `round(100 * 0 / (0 + 0))`.

essen:
Nice, thanks for the report!

hdima:
A quick fix is to replace `round(100 * Y / (Y + N))` with `case Y + N of 0 -> 0; S -> round(100 * Y / S) end`.

essen:
Probably going to be the one merged, can you send a PR?

hdima:
OK, I'll send a PR little bit later today

hdima:
To add little bit of confusion, we discussed it in the team and some people said that the final coverage for the module without any functions should be 100% (instead of 0%).

essen:
Makes sense.

essen:
Closing as the PR has long been merged. Thanks again!

PR #410, opened 2015-10-23
Fix possible 'badarith' error for cover

hdima:
Fix possible division by zero in case if a module doesn't export any functions. See #409 for more details.


essen:
Merged, thanks!

PR #411, opened 2015-10-26
deps.mk: In dep_target, call dep_autopatch during the second eval

dumbbell:
... not the first.

`dep_autopatch` can be an expensive macro, in particular because it executes Erlang. So if `NO_AUTOPATCH` is set, it's best to avoid calling `dep_autopatch` as the result will be unused anyway.

### DIFF
--- a/core/deps.mk
+++ b/core/deps.mk
@@ -179,7 +179,7 @@ define dep_autopatch_rebar.erl
 		file:write_file("$(call core_native_path,$(DEPS_DIR)/$1/Makefile)", Text, [append])
 	end,
 	Escape = fun (Text) ->
-		re:replace(Text, "\\\\$$$$", "\$$$$$$$$", [global, {return, list}])
+		re:replace(Text, "\\\\$$", "\$$$$", [global, {return, list}])
 	end,
 	Write("IGNORE_DEPS += edown eper eunit_formatters meck node_package "
 		"rebar_lock_deps_plugin rebar_vsn_plugin reltool_util\n"),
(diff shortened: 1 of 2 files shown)


essen:
Agree in theory but the $$(...) form confuses me. What's the more human readable equivalent? :-)

dumbbell:
Some numbers while I'm at it: in the case of RabbitMQ, a no-op make dep went down from 21 seconds to 5 seconds.

dumbbell:
Another possibility is to use something like `$(if <test NO_AUTOPATCH>,$(call dep_autopatch))`, but I'm not sure it's more human friendly.

ash-lshift:
This breaks the `dep_autopatch_rebar.erl` target of `erlang.mk`.
To reproduce the problem, depend on a package which uses rebar and `port_specs`, eg https://github.com/vinoski/erlsha2 or https://github.com/davisp/jiffy, then run `make deps` and you will find that it fails because the auto-patching has corrupted `deps/$DEP/Makefile` by writing apparently random numbers into it.
The string `$$$$` appears several times in `dep_autopatch_rebar.erl` and I think this patch causes `$$$$` to be interpolated into a PID.

essen:
Thanks for the info. Looks like more work is needed then. :-)

essen:
@dumbbell So basically, the fact that this part is inside an ifeq ... that results to not running the content still causes the insides of the ifeq to be expanded? If true, didn't know that.

dumbbell:
Here is a new patch.

@ash-lshift: I added a test to confirm the problem you reported. The `dep_autopath_rebar.erl` macro needed one less level of `$`-escaping.

essen:
Looks fine, will experiment with this ticket next.

essen:
It looked good to me when I experimented the other day, but just to make sure I want to check against all packages. I will try to do two full runs, one before, one after, and see if nothing breaks.

essen:
Before this patch, the applications mu2, smother and wrangler were giving a load error. After this patch, they don't. Something changed, not sure what yet.

essen:
Figured out the difference. There was a $$$$ -> $$ conversion that you forgot (the one for .app handling in https://github.com/rabbitmq/erlang.mk/blob/late-autopatch/core/deps.mk#L472). Can you add it to your commit? Then I can merge.

dumbbell:
Thanks for debugging this! I updated the branch.

essen:
Great, meanwhile I tested again with this change against all packages and everything checks out. Will merge in a minute.

essen:
Merged! Thanks.

PR #412, opened 2015-10-26
V=2 causes `set -x` to be set on forked shells

dumbbell:
This helps to debug some recipes which are more than one line of shell script, for instance when the recipe is a shell `for` loop.

### DIFF
--- a/test/Makefile
+++ b/test/Makefile
@@ -82,7 +82,7 @@ else ifeq ($V,2)
 	i = @echo == $@:
 else
 	t =
-	v = V=1
+	v = V=$(shell echo $$(($(V)-2)))
 	i = @echo == $@:
 endif
 
--- a/core/core.mk
+++ b/core/core.mk
@@ -30,9 +30,11 @@ PROJECT_VERSION ?= rolling
 V ?= 0
 
 verbose_0 = @
+verbose_2 = set -x;
 verbose = $(verbose_$(V))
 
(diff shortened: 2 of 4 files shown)


essen:
Does this mean that, for example, "set -x; make" won't work?

dumbbell:
Yeah, `set -x; make` does not achieve the same result: it sets `-x` on the current shell, not on the shells spawned by make(1). It's possible that some Bourne shell implementations somehow pass this flag to child shells, I have no idea.

test/Makefile
```diff
@@ -82,7 +82,7 @@ else ifeq ($V,2)
 	i = @echo == $@:
 else
 	t =
-	v = V=1
+	v = V=$(shell echo $$(($(V)-2)))
```
essen: Looks fine, only one question about this one here: what does it do?

dumbbell: The goal is to compute the value of V:
* `make check V=3` gives `make V=1`
* `make check V=4` gives `make V=2`

essen: OK found my way to some relevant documentation. Had no idea shell arithmetic was a thing. I'll drop this link for future reference: http://bash.cyberciti.biz/guide/Perform_arithmetic_operations

Looks good then. There's a comment that needs updating above this, but I'll do it while merging.

essen:
Merged, thanks!

PR #413, opened 2015-10-27
Use an intermediate timestamp file to know when to touch source files

dumbbell:
The previous approach spawned a shell for every single source files. As we disable parallel make jobs, this is very time consuming.

The new approach uses an intermediate timestamp file in `$(ERLANG_MK_TMP)` to record the last modification of any of `$(MAKEFILE_LIST)` and touch all source files in one command. Then, the `.app` file depends on this timestamp file.

We test the existence of this timestamp file: if if doesn't exist, don't touch source files, they will be built anyway.

### DIFF
--- a/core/erlc.mk
+++ b/core/erlc.mk
@@ -216,13 +216,23 @@ define makedep.erl
 endef
 
 ifeq ($(if $(NO_MAKEDEP),$(wildcard $(PROJECT).d),),)
-$(PROJECT).d:: $(ERL_FILES) $(call core_find,include/,*.hrl)
+$(PROJECT).d:: $(ERL_FILES) $(call core_find,include/,*.hrl) $(MAKEFILE_LIST)
 	$(makedep_verbose) $(call erlang,$(call makedep.erl,$@))
 endif
 
+ifneq ($(words $(ERL_FILES) $(CORE_FILES) $(ASN1_FILES) $(MIB_FILES) $(XRL_FILES) $(YRL_FILES)),0)
 # Rebuild everything when the Makefile changes.
-$(ERL_FILES) $(CORE_FILES) $(ASN1_FILES) $(MIB_FILES) $(XRL_FILES) $(YRL_FILES):: $(MAKEFILE_LIST)
+$(ERLANG_MK_TMP)/last-makefile-change: $(MAKEFILE_LIST)
+	@mkdir -p $(ERLANG_MK_TMP)
+	@if test -f $@; then \
+		touch $(ERL_FILES) $(CORE_FILES) $(ASN1_FILES) $(MIB_FILES) $(XRL_FILES) $(YRL_FILES); \
+		touch -c $(PROJECT).d; \
+	fi
 	@touch $@
+
+$(ERL_FILES) $(CORE_FILES) $(ASN1_FILES) $(MIB_FILES) $(XRL_FILES) $(YRL_FILES):: $(ERLANG_MK_TMP)/last-makefile-change
+ebin/$(PROJECT).app:: $(ERLANG_MK_TMP)/last-makefile-change
+endif
 
 -include $(PROJECT).d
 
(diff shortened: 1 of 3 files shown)


dumbbell:
Again, some numbers: in the case of RabbitMQ, a fresh build of the full distribution went from 1:23 to down 0:49.

essen:
I think we are quite close to be able to reenable parallel so I will wait for that to be done before I consider merging this PR.

essen:
Working on this one now, or rather, on enabling parallel again. A tiny change makes almost all tests pass. I suspect there's issues with releases however (lacking a few more complex tests) and it seems the ErlyDTL plugin doesn't like it yet. These do not concern RabbitMQ however so I will compare and get back to you.

essen:
First test. Cloning the umbrella and then running make with/without parallel:

make  26.55s user 6.42s system 11% cpu 4:45.32 total
make -j 32  25.35s user 6.30s system 20% cpu 2:36.91 total

Next I will try building.

dumbbell:
RabbitMQ's erlang.mk plugins will need fixes as well.

essen:
Actually about the above test, I only changed the umbrella's Erlang.mk file so far, so it probably will be a bigger increase as well.

essen:
Not seeing much improvement building stuff. I'll finish that and let you run the make packages thing.

essen:
Moving the ErlyDTL plugin above erlc in build.config fixed the issues. I suppose I can disable parallel when building a release and leave it enabled otherwise.

core/erlc.mk
```diff
@@ -179,14 +179,23 @@ define makedep.erl
...
 ifeq ($(if $(NO_MAKEDEP),$(wildcard $(PROJECT).d),),)
-$(PROJECT).d:: $(ERL_FILES) $(call core_find,include/,*.hrl)
+$(PROJECT).d:: $(ERL_FILES) $(call core_find,include/,*.hrl) $(MAKEFILE_LIST)
 	$(makedep_verbose) $(call erlang,$(call makedep.erl,$@))
 endif
 
+ifneq ($(words $(ERL_FILES) $(CORE_FILES) $(ASN1_FILES) $(MIB_FILES) $(XRL_FILES) $(YRL_FILES)),0)
 # Rebuild everything when the Makefile changes.
-$(ERL_FILES) $(CORE_FILES) $(ASN1_FILES) $(MIB_FILES) $(XRL_FILES) $(YRL_FILES):: $(MAKEFILE_LIST)
+$(ERLANG_MK_TMP)/last-makefile-change: $(MAKEFILE_LIST)
```
essen: I do not understand why this refers to a file that apparently never exists. I'm missing something here.

core/erlc.mk
```diff
@@ -179,14 +179,23 @@ define makedep.erl
...
+	@mkdir -p $(ERLANG_MK_TMP)
+	@if test -f $@; then \
+		touch $(ERL_FILES) $(CORE_FILES) $(ASN1_FILES) $(MIB_FILES) $(XRL_FILES) $(YRL_FILES); \
+		touch -c $(PROJECT).d; \
+	fi
 	@touch $@
```
essen: And I don't understand what this does, it sounds like all files are touched first and then some files are touched again? I'm lost.

essen: Oooooh this is for the last-makefile-change, nevermind then I understand all this.

essen:
Now that parallel is enabled again, I believe this patch should be included only if it provides a significant improvement over a parallel build, to avoid adding extra complexity.

essen:
CI started.

essen:
Tests look fine.

essen:
Only one thing needed before merge: replace `@` by `$(verbose)` to silence commands, as the latter can be deactivated with `V=1`.

essen:
Merged, thanks!

PR #361, opened 2015-09-12
Add possibility to specify custom ErlyDTL options

KrzysiekJ:

PR #419, opened 2015-11-13
Fix Erlydtl, add DTL_OPTS

jonasrichard:
There was a misplaced ) which resulted that erl didn't get ebin as -pa parameter. So my custom filter was never picked up.
I added DTL_OPTS in order that one can add

```
[{libraries, [{mylib, mylib_module}]}, {default_libraries, [mylib]}]
```

easily.

### DIFF
--- a/plugins/erlydtl.mk
+++ b/plugins/erlydtl.mk
@@ -4,6 +4,7 @@
 # Configuration.
 
 DTL_FULL_PATH ?=
+DTL_OPTS ?= []
 DTL_PATH ?= templates/
 DTL_SUFFIX ?= _dtl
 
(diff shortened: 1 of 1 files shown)


jonasrichard:
Builds are failed but they are probably independent from this fix.

essen:
Yes it's very likely to be unrelated currently.

essen:
I have fixed the paren in the most recent commit (to fix related issues). I have not reviewed the rest yet.

essen:
I'm all for a DTL_OPTS but this needs tests.

essen:
Closing in favor of https://github.com/ninenines/erlang.mk/pull/361

PR #423, opened 2015-11-17
make erlang-mk: Copy build.config after switching to $(ERLANG_MK_COMMIT)

dumbbell:
If the local build.config has changes compared to upstream's master branch, and `$(ERLANG_MK_COMMIT)` also brings changes to this file, copying the local file to the checkout before switching to `$(ERLANG_MK_COMMIT)` caused git-checkout(1) to abort with:

```
error: Your local changes to the following files would be overwritten by checkout:
	build.config
Please, commit your changes or stash them before you can switch branches.
Aborting
```

### DIFF
--- a/core/core.mk
+++ b/core/core.mk
@@ -192,8 +192,11 @@ ERLANG_MK_BUILD_DIR ?= .erlang.mk.build
 
 erlang-mk:
 	git clone $(ERLANG_MK_REPO) $(ERLANG_MK_BUILD_DIR)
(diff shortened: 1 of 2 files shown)


essen:
Would love to have a test for that one to make sure it doesn't happen again.

dumbbell:
I added a test.

The patch also contains a fix to the `core-upgrade-custom-repo` test case; see commit message.

essen:
Merged, thanks!

PR #424, opened 2015-11-18
adds french accent to rm

videlalvaro:


essen:
I will be closing this without merging.

While I think the idea is very neat and it would bring a lot to the project, I don't see a way to enforce this easily in the future, especially considering most people's habits seem to favor the current form. We will need to find a better way to add French accents to the project, one that doesn't require constant attention for all future changes.

Thanks for the proposal though!

PR #425, opened 2015-11-18
Do not export templates as environment variables

dumbbell:
Use `$(file)` macro to write the generated files. This avoids environment pollution with a large amount of Erlang code.

### DIFF
--- a/plugins/bootstrap.mk
+++ b/plugins/bootstrap.mk
@@ -344,7 +344,7 @@ endef
 # Plugin-specific targets.
 
 define render_template
-	$(verbose) echo "$${_$(1)}" > $(2)
+	$(verbose) printf -- '$(subst $(newline),\n,$(subst %,%%,$(subst ','\'',$(subst $(tab),$(WS),$(call $(1))))))\n' > $(2)
 endef
 
 ifndef WS
@@ -354,10 +354,6 @@ else
 WS = $(tab)
 endif
 endif
-
-$(foreach template,$(filter bs_% tpl_%,$(.VARIABLES)), \
-	$(eval _$(template) = $$(subst $$(tab),$$(WS),$$($(template)))) \
-	$(eval export _$(template)))
 
 bootstrap:
 ifneq ($(wildcard src/),)
--- a/plugins/c_src.mk
+++ b/plugins/c_src.mk
@@ -192,10 +192,6 @@ hello(_) ->
 	erlang:nif_error({not_loaded, ?MODULE}).
 endef
 
-$(foreach template,bs_c_nif bs_erl_nif, \
-	$(eval _$(template) = $$(subst $$(tab),$$(WS),$$($(template)))) \
-	$(eval export _$(template)))
-
 new-nif:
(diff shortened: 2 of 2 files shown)


essen:
We can't use $(file) because that's Make 4+ (or 4.1, don't remember exactly) and we still need 3.81+ for stock OSX.

dumbbell:
I modified the patch to use printf(1) instead of `$(file)` and tested it with GNU Make 3.81.

plugins/bootstrap.mk
```diff
@@ -366,7 +354,7 @@ endif
 	$(eval p := $(PROJECT))
 	$(eval n := $(PROJECT)_sup)
 	$(call render_template,bs_Makefile,Makefile)
-	$(verbose) mkdir src/
+	$(verbose) mkdir -p src/
```
essen: This change is not needed, since just before that we check that src/ doesn't exist and error out otherwise.

plugins/bootstrap.mk
```diff
@@ -379,7 +367,7 @@ ifneq ($(wildcard src/),)
 endif
 	$(eval p := $(PROJECT))
 	$(call render_template,bs_Makefile,Makefile)
-	$(verbose) mkdir src/
+	$(verbose) mkdir -p src/
```
essen: Same here.

plugins/bootstrap.mk
```diff
@@ -393,7 +381,7 @@ ifneq ($(wildcard rel/),)
 endif
 	$(eval p := $(PROJECT))
 	$(call render_template,bs_relx_config,relx.config)
-	$(verbose) mkdir rel/
+	$(verbose) mkdir -p rel/
```
essen: Same here, except with rel/.

dumbbell:
Ok, fixed. It was required with the previous iteration but not anymore.

plugins/bootstrap.mk
```diff
@@ -344,21 +344,9 @@ endef
...
+	$(verbose) printf -- '$(subst $(newline),\n,$(subst %,%%,$(subst ','\'',$(call $(1)))))\n' > $(2)
 endef
 
-ifndef WS
-ifdef SP
-WS = $(subst a,,a $(wordlist 1,$(SP),a a a a a a a a a a a a a a a a a a a a))
-else
-WS = $(tab)
-endif
-endif
```
essen: We still need this (and the subst below), you are removing the feature that allows users to define what identation to use in bootstrapped/generated files by setting SP to the number of spaces per indentation (or tabs if SP is undefined).

essen:
Again don't worry about tests, this time I used this ticket to trigger a bug in BuildKite and report it. But the above issue is real. :-)

dumbbell:
I didn't know about this feature. I restored it.

essen:
Yeah, I suppose it's not tested, and possibly not documented, so I will fix this after merging. Testing now.

essen:
So this is OK but I will only push the merge after I add the missing test (probably tomorrow).

essen:
Merged, thanks!

PR #437, opened 2015-11-28
Fix for bcrypt compilation on OSX El Capitan

elbrujohalcon:


essen:
Quick look confirms this works. Doing a full run with your patch now. Can you make it a single commit?

essen:
Full run worked and also fixed erlpass and ierlang. Just need a good squash now.

elbrujohalcon:
sqashed :+1:

essen:
I have a last full run currently running, this time on Linux, and then it should be good.

essen:
Merged, thanks!

PR #445, opened 2015-12-09
Improve the customisation of edoc, with default values

jeanparpaillon:
edoc options are set through:
- default values (proplists): 
  - {subpackages, false}
  - {source_path, ALL_APPS_DIRS + ALL_DEPS_DIRS}
- 'source_path' can be overriden with EDOC_SRC_DIRS
- all options can be overriden with EDOC_OPTS

### DIFF
--- a/plugins/edoc.mk
+++ b/plugins/edoc.mk
@@ -11,7 +11,9 @@ EDOC_SRC_DIRS ?= $(ALL_APPS_DIRS) $(ALL_DEPS_DIRS)
 # TODO: use double-quote instead of single + atom_to_list
 #       The problem is in correctly escaping double-quotes
 define edoc.erl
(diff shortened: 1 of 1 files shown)


essen:
Looks sensible, just one question. Why put single quotes and then convert to list? If there's a good reason, should put a comment above the .erl block.

jeanparpaillon:
Just updated the pr with a comment on why using single quotes instead of doubles.
Basically, a problem of double-quote escaping. There may be a solution but I have not found it ;)

essen:
Not quite sure about `$(ALL_DEPS_DIRS)`. Also, this needs tests. :-)

essen:
I have merged and did a few tweaks after, and added a test. I don't think there should be any path by default so I removed them. Thanks!

Issue #370, opened 2015-09-24
Adding a cheat sheet to the docs


nyaray:
It would be immensely helpful to have a cheat sheet with useful (and sometimes rarely executed) commands. I'm thinking something along the lines of the following, but don't really know where it could/should/would go, so I'm just filing this here.

# erlang.mk Cheat Sheet (starting point)

This document provides some useful (and maybe less commonly executed) commands.

## _"How Do I"..._

Issue | Command
----|----
See all available targets | `make help`
Update erlang.mk | `make erlang-mk`
Build a project | `make`
Fetch and build dependencies | `make deps`
[TO BE DECIDED] Update the dependencies | `make ...`
Run a release | `make run`
List packages | `make search`
Search across all packages | `make search q=json`
List available templates | `make list-templates`
Generate a `gen_server` module | `make new t=gen_server n=my_server`

### Configuration

Adding dependencies is done by adding package names to the `DEPS` variable in `Makefile` before the `include erlang.mk` line. List or search in the packages as described above to find something you might need. An example `Makefile`:

    PROJECT = "my_project"
    DEPS = cowboy
    include erlang.mk

## Getting started

Run the following commands, changing `PROJECT_DIR` as you see fit:

    PROJECT_DIR="project"; mkdir $PROJECT_DIR && cd $PROJECT_DIR\
    URL="https://raw.githubusercontent.com/ninenines/erlang.mk/master/erlang.mk"; if [ -n $(command -v wget) ]; then; wget $URL; else; curl $URL; fi;

## Bootstrapping

OTP Pattern | Command
---|---
Application | `make -f erlang.mk bootstrap`
Library | `make -f erlang.mk bootstrap-lib`
Release | `make bootstrap-rel` (can be appended to any of the previous targets)

Bootstrap for the kind of project you want to create. If you want to build a release containing an application or a library you can combine either target with `bootstrap-rel`. Bootstrapping a release running an OTP application:

    make -f erlang.mk bootstrap bootstrap-rel

__NOTE:__ When you've used bootstrap-rel `make` will cause the release to be built.

Run the built release:

    make run

nyaray:
Oops, forgot to add `make shell` somewhere in the "How Do I..." section.

PR #468, opened 2015-12-29
Fix invalid link to documentation

### LINKED ISSUE #467: Missing link in app.asciidoc
https://github.com/ninenines/erlang.mk/blob/master/doc/src/guide/relx.asciidoc

for memory, too

crownedgrouse:
fix #467

### DIFF
--- a/doc/src/guide/app.asciidoc
+++ b/doc/src/guide/app.asciidoc
@@ -94,7 +94,7 @@ the application and all dependencies:
 [source,bash]
 $ make rel
 
-Consult the link:relx.asciidoc[Releases] chapter for more
+Consult the link:releases.asciidoc[Releases] chapter for more
 information about what releases are and how they are generated.
 
 === Application resource file


essen:
Merged, thanks!

essen:
Actually I most likely will broke it again very soon, and won't be able to fix it. This is due to the substandard asciidoc support on Github. You should switch to http://erlang.mk for reading the documentation (links are broken THERE because of a change I did yesterday, I am fixing it now and this will break github).

crownedgrouse:
OK, so what about using a minimal Wiki on github with links to http://erlang.mk , instead ? And a simple cheat sheet in README.md like #370 is proposing may be good. MHO

crownedgrouse:
I precise my thoughts : wiki on github for experimental/master features and a link to http://erlang.mk for things written in marble / official stable documentation.

essen:
There's no difference between master and stable in this project. Documenting experimental features on the wiki would still make them documented and that's not something we want. There's very little of those anyway. Finally there's already a link to erlang.mk at the top of the repository and to the guide in the README.

essen:
Ah, I see, README is still old links.

crownedgrouse:
yes, this is why people, like me, are looking at documentation in github instead of http://erlang.mk

crownedgrouse:
cheat sheet would be nice anyway, and a heading global link to http://erlang.mk documentation would be sufficient , instead many 'Check out our upcoming user guide!' ... but a personal feeling.

essen:
The documentation is slowly phasing out the old README, which explains why things are the way they are now. But it should be time to rework it now. I will do it today.

A cheat sheet would be nice, and will be added at some point. `make help` is the best alternative for now.

PR #475, opened 2016-01-05
Fix .app compilation when a git tag contains a '/'

jeanparpaillon:
When the last git tag contains '/' character, compilation of .app file fails because it interfers with sed regexp syntax


essen:
Nice find, thanks.

Do it where sed is used, not here, because when there's no .app.src file Erlang.mk doesn't use sed to create the .app file.

essen:
Merged, thanks!

PR #483, opened 2016-01-07
Create ebin directory if it doesn't exist before erlydtl compile.

hairyhum:
Maybe there are better way to do so. In some make task ordering? Or make erlang compile phase run before erlydtl.

### DIFF
--- a/plugins/erlydtl.mk
+++ b/plugins/erlydtl.mk
@@ -38,6 +38,11 @@ define erlydtl_compile.erl
 				re:replace(F2, "/",  "_",  [{return, list}, global])
 		end,
 		Module = list_to_atom(string:to_lower(Module0) ++ "$(DTL_SUFFIX)"),
+		case file:make_dir("ebin/") of
+			ok -> ok;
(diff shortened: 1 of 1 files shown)


essen:
This is already done here: https://github.com/ninenines/erlang.mk/pull/483/files#diff-8af0c03a140a70f06e62bfe8e3d1361aR54

essen:
What's the exact issue you have, and is this patch in your version of erlang.mk?

hairyhum:
I'm trying to build app with .dtl template in `template/` folder while not having `ebin/` folder (after `make clean`).
And it fails with `{{case_clause,{error,[{"templates/auth_form.dtl",[{none,erlydtl_beam_compiler,{write_file,enoent}}]}],[]}}`
If I create `ebin/` folder it doesn't fail.
I have updated erlang.mk with make erlang-mk, maybe I didn't have latest version?

hairyhum:
Sorry, looks like rabbitmq fork is outdated and doesn't have this fix.

essen:
Cheers.

PR #500, opened 2016-01-28
[Fix #497] Support for hex package names

### LINKED ISSUE #497: Support for hex.pm packages
In `hex` fetcher, allow users to specify the name of the package (which can be different than the associated application):

In the following example `aleppo` is the name of the application and `inaka_aleppo` the package:

```
DEPS = aleppo
dep_aleppo = hex 0.9.9 inaka_aleppo 
```

elbrujohalcon:
[Fix #497] Support for hex package names

### DIFF
--- a/core/deps.mk
+++ b/core/deps.mk
@@ -481,8 +481,13 @@ endef
 define dep_fetch_hex.erl
 	ssl:start(),
 	inets:start(),
+	Url =
+		case "$(3)" of
+			"" -> "https://s3.amazonaws.com/s3.hex.pm/tarballs/$(1)-$(2).tar";
+			_ -> "https://s3.amazonaws.com/s3.hex.pm/tarballs/$(3)-$(2).tar"
+		end,
 	{ok, {{_, 200, _}, _, Body}} = httpc:request(get,
-		{"https://s3.amazonaws.com/s3.hex.pm/tarballs/$(1)-$(2).tar", []},
+		{Url, []},
 		[], [{body_format, binary}]),
 	{ok, Files} = erl_tar:extract({binary, Body}, [memory]),
 	{_, Source} = lists:keyfind("contents.tar.gz", 1, Files),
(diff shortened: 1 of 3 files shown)


elbrujohalcon:
@essen I'm not sure those errors are produced by my changes… are they?

essen:
Probably not. If "make check c=core" works for you, don't worry.

elbrujohalcon:
@essen `make check c=core` blocks forever at…
```
core-deps-fetch-svn: Bootstrap a new OTP library named test_core_deps_fetch_svn
core-deps-fetch-svn: Add Cowlib 1.0.0 to the list of dependencies
core-deps-fetch-svn: Build the application
```
…any clue on what can I be doing wrong here?

elbrujohalcon:
@essen it works! I took the chance and added tests for the new functionality :)

elbrujohalcon:
@essen any plans on how this will be merged? Anything else I need to do for that?

essen:
Just need to squash I think.

I'll try to do a merge run at the end of the week. Can't promise anything right now, personal stuff.

elbrujohalcon:
@essen squashed, it is.

elbrujohalcon:
@essen : after-erlang-factory ping :trollface:

essen:
I will dedicate one day to merging user contributions this week, but today is not the day.

essen:
So as mentioned on #529 I think first we need to have the == hardcoded and on position 2, and then this new argument would be on position 4.

essen:
@elbrujohalcon @glejeune I think the Hex related development would progress a lot faster if it happened in the hexpm plugin. I propose to do the following for the time being:

* Remove the hex functionality from Erlang.mk (except autopatch, see next point)
* Autoload the hexpm plugin when a fetch method "hex" is detected
* Add a warning when the 2-word form of "hex" is detected (see below).
* (I think a warning is also in order when the older "git" forms are detected.)

This change would avoid duplicating development between hex/hexpm. On the other hand the plugin has to support the "hex" fetch method. I recommend making hex exactly like hexpm (maybe an alias) except when the dep_$1 line has only 2 words (example "hex 1.0") where "==" is implied. This is only for compatibility and could be removed in a few months. The warning would speed up the change.

From that point onward, it's a wait and see game. Development wouldn't be hampered by my busy schedule, and we can later decide if it should stay a pure plugin or should be part of the core.

Thoughts?

essen:
By the way in addition to what I said, I would also change autopatch to use "hex == $(VERSION)", so the 2-words "hex" stuff wouldn't be there for long.

elbrujohalcon:
This PR is a bit old, I think I'll just close it :trollface:

PR #504, opened 2016-02-01
Allow to generate .app.src

jeanparpaillon:
Dependancy between .app and .app.src file is calculated when loading Makefile, because of the use of 'wildcard' function.

This patch allows to generate .app.src file, adding a double-colon rule on .app.src, for instance:
```make
src/$(PROJECT).app.src: my_custom_dep
```

Auto-generated .app file has been tested also.

### DIFF
--- a/core/erlc.mk
+++ b/core/erlc.mk
@@ -234,24 +234,29 @@ define compile_erl
 		-pa ebin/ -I include/ $(filter-out $(ERLC_EXCLUDE_PATHS),$(COMPILE_FIRST_PATHS) $(1))
 endef
 
-ebin/$(PROJECT).app:: $(ERL_FILES) $(CORE_FILES) $(wildcard src/$(PROJECT).app.src)
+ebin/$(PROJECT).app:: $(ERL_FILES) $(CORE_FILES) src/$(PROJECT).app.src
 	$(eval FILES_TO_COMPILE := $(filter-out src/$(PROJECT).app.src,$?))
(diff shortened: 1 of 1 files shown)


essen:
I don't understand what this patch is doing.

jeanparpaillon:
Actually, if you write the following rule to generate .app.src, for instance:
```make
src/$(PROJECT).app.src: src/$(PROJECT).app.src.in
     gen $< > $@
```

It does not work because `ebin/$(PROJECT).app` target is generated before the custom rule is executed. 
Reason: dependancy between `ebin/$(PROJECT).app` and `src/$(PROJECT).app.src` is calculated with `wildcard` function.

My patch makes `ebin/$(PROJECT).app` _always_ depends on `src/$(PROJECT).app.src`, which is a fake rule if .app is auto-generated.

jeanparpaillon:
I've updated the patch, was previously broken, sorry

essen:
From what I understand you should have no problem with generating the .app.src if you were doing it as part of `$(PROJECT).d::` because the `ebin/$(PROJECT).app::` target is run in a separate make invocation and therefore the .app.src will exist there. Did you try that?

jeanparpaillon:
Right, my bad ...

essen:
No worries. :-) Do tell if you get it working this way though.

jeanparpaillon:
I confirm it works with dependancy on $(PROJECT).d

essen:
Cheers!

PR #507, opened 2016-02-03
[docs] various doc improvements (multi-apps, doc requirements)

ingwinlu:


essen:
As I said on IRC, xsltproc is also a requirement (you probably had it already, it's part of libxslt usually). And the last requirement I forgot is LaTeX itself, it's not always installed with dblatex.

ingwinlu:
dblatex lists xsltproc and latex as a dependency, so I see no need of including it in the listing of erlang.mk's dependencies (http://dblatex.sourceforge.net/ "Dependencies").

essen:
But it's not always installed as a dependency, on some systems you have to install it separately.

ingwinlu:
xsltproc sometimes is not handled as an dependency (more as an optional dependency) because you can also use different XSLT's. So to write that you need xsltproc would actually be wrong.

To build the docs you just need a working installation of asciidoc and dblatex. How to get those two requirements to work is not in the scope of erlang.mk.

essen:
a2x (asciidoc to xml, http://asciidoc.org/a2x.1.html) used by Erlang.mk does require xsltproc(1). For Erlang.mk itself you need the first four in this list, at least for now: http://asciidoc.org/a2x.1.html#X1 (I had forgotten about the XSL stylesheets too).

essen:
Note that I use (or will use, if I don't yet) the --xsltproc-opts option for a2x and that one implies you put xsltproc(1) options. I don't know if the alternatives have exactly the same options, but if not, then we definitely depend on xsltproc.

ingwinlu:
I see; updated.

essen:
Merged, thanks!

PR #511, opened 2016-02-08
Fix ct for multi-app repositories

ingwinlu:
* Include $(APPS_DIR)/*/ebin in CT_RUN
* test-build before ct-apps
* Test that my_lib is accessible from my-app ct

Currently ct is not really working for multi-application repositories. This PR addresses and fixes the issue though further restructuring of the CT plugin might be advisable in the future.

Local test run failed on dialyzer tests that I think are unrelated.


essen:
Merged, thanks!

PR #518, opened 2016-02-16
Add package jsone to the index

antik486:
**jsone**
An Erlang library for encoding, decoding JSON data.
https://github.com/sile/jsone


essen:
Merged, thanks!

PR #519, opened 2016-02-16
Old repository is no longer actively maintained

antik486:
Old repository is no longer actively maintained

https://github.com/klarna/jesse/blob/master/README.md


essen:
Merged, thanks!

PR #520, opened 2016-02-17
Add empty 'module' template

ingwinlu:
Every .erl source file will need at least a -module and a single -export
definition. With the empty template a file can be quickly created that
already populates -module with the right value as well as an empty
export field that can quickly be filled.

### DIFF
--- a/plugins/bootstrap.mk
+++ b/plugins/bootstrap.mk
@@ -172,6 +172,11 @@ code_change(_OldVsn, State, _Extra) ->
 	{ok, State}.
 endef
 
+define tpl_module
+-module($(n)).
+-export([]).
+endef
+
 define tpl_cowboy_http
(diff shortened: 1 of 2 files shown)


ingwinlu:
maybe 'blank' would have been a better name :/

essen:
Thought we already had. The name is not clear as you point out. I'm not sure what a good name would be. I don't think 'blank' is good either as it wouldn't come to my mind at least. Maybe we can call it 'module'. If writing about creating an empty module I would say something like "just a module" as opposed to a gen_server or whatnot. Thoughts?

ingwinlu:
module sounds reasonable, updated

essen:
Cheers. Looks good. I'll let the tests run and merge later.

essen:
Merged, thanks!

PR #521, opened 2016-02-18
escript file maybe different from escript name

zmstone:


essen:
Looks sensible enough. Letting tests run and will merge this and a few PRs at the same time probably next week or so.

essen:
Thanks!

essen:
Merged, thanks!

PR #524, opened 2016-02-21
Add the Mix plugin to the list

glejeune:


essen:
Merged and should be published to erlang.mk, thanks!

PR #535, opened 2016-03-17
Do not use grep Extended Mode

### LINKED ISSUE #523: BusyBox/Alpine grep -E causes error on app-build target
Attempting to build ranch on Alpine linux results in the following error:

```
grep: bad regex '^[^%]*{\s*modules\s*,': Invalid contents of {}
Empty modules entry not found in ranch.app.src. Please consult the erlang.mk README for instructions.
erlang.mk:4898: recipe for target 'ebin/ranch.app' failed
```

This appears to be caused by the -E option on https://github.com/ninenines/ranch/blob/master/erlang.mk#L4907

```
bash-4.3# grep -E '^[^%]*{\s*modules\s*,' src/ranch.app.src
grep: bad regex '^[^%]*{\s*modules\s*,': Invalid contents of {}
```

If I try the regex in the shell without the extended option it works:

```
bash-4.3# grep '^[^%]*{\s*modules\s*,' src/ranch.app.src
        {modules, []},
```

For reference:
```
bash-4.3# grep -v
BusyBox v1.24.1 (2015-12-16 08:00:02 GMT) multi-call binary.
```

I'm attempting to compile and package an elixir app using the Elixir Docker Image Builder (https://github.com/edib-tool/elixir-docker-image-builder). It is based on Alpine 3.3 which is where the grep binary seems to be coming from.

Thanks,
Joel

ingwinlu:
Extended Mode uses { and } as special characters. Making grep
implementations other then GNU grep fail (see
http://www.gnu.org/software/grep/manual/grep.html#Basic-vs-Extended).

---

I assume -E is a typo for -e as Extended Mode is not needed for that grep.
Resolves #523.

### DIFF
--- a/core/erlc.mk
+++ b/core/erlc.mk
@@ -244,7 +244,7 @@ ifeq ($(wildcard src/$(PROJECT).app.src),)
 	$(app_verbose) printf "$(subst $(newline),\n,$(subst ",\",$(call app_file,$(GITDESCRIBE),$(MODULES))))" \
 		> ebin/$(PROJECT).app
 else
-	$(verbose) if [ -z "$$(grep -E '^[^%]*{\s*modules\s*,' src/$(PROJECT).app.src)" ]; then \
+	$(verbose) if [ -z "$$(grep -e '^[^%]*{\s*modules\s*,' src/$(PROJECT).app.src)" ]; then \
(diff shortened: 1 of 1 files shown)


essen:
NIce thanks.

essen:
Merged, thanks!

PR #543, opened 2016-04-03
add WITHOUT variable to the makefile

### LINKED ISSUE #537: lighter erlang.mk?
This is quite an open question, but I am wondering why the bootstrapped `erlang.mk` should contain a default list of package, where it could only contains the "build logic"?

Indeed I can understand the usefulness of having  a list of default packages but I really wish that my projects don't have to embed all these links to unused package when I ship them. Instead why not having them available in a separate repository and have a tool to fetch infos about them and generate a file that could then be included or used by erlang.mk?

Something like the [carthage file](https://github.com/Carthage/Carthage/blob/master/Documentation/Artifacts.md#cartfile) 

Thoughts?

benoitc:
This allows to ignore lines from the default build.config if needed.
For example to not include packages run make WITHOUT=index

fix #537

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -12,8 +12,16 @@
 # ACTION OF CONTRACT, NEGLIGENCE OR OTHER TORTIOUS ACTION, ARISING OUT OF
 # OR IN CONNECTION WITH THE USE OR PERFORMANCE OF THIS SOFTWARE.
 
+WITHOUT ?=
+
 BUILD_CONFIG_FILE ?= $(CURDIR)/build.config
+ifeq ($(WITHOUT),)
 BUILD_CONFIG = $(shell sed "s/\#.*//" $(BUILD_CONFIG_FILE))
+else
+empty := $(subst ,, )
+BUILD_EXCLUDE = "^$(subst $(empty),\|^,$(WITHOUT))"
+BUILD_CONFIG = $(shell sed "s/\#.*//" $(BUILD_CONFIG_FILE)|grep -v $(BUILD_EXCLUDE))
+endif
 
 ERLANG_MK = erlang.mk
 ERLANG_MK_VERSION = $(shell git describe --tags --dirty)
--- a/Makefile
+++ b/Makefile
@@ -31,7 +31,7 @@ ERLANG_MK_VERSION = $(shell git describe --tags --dirty)
 all:
 	awk 'FNR==1 && NR!=1{print ""}1' $(patsubst %,%.mk,$(BUILD_CONFIG)) \
 		| sed 's/^ERLANG_MK_VERSION = .*/ERLANG_MK_VERSION = $(ERLANG_MK_VERSION)/' \
-		| sed 's/^DEFAULT_WITHOUT = .*/DEFAULT_WITHOUT = $(WITHOUT)/' > $(ERLANG_MK)
+		| sed 's:^DEFAULT_WITHOUT = .*:DEFAULT_WITHOUT = $(WITHOUT):' > $(ERLANG_MK)
 
 ifdef p
 # Remove p from the list of variables since that conflicts with bootstrapping.
--- a/Makefile
+++ b/Makefile
@@ -31,7 +31,7 @@ ERLANG_MK_VERSION = $(shell git describe --tags --dirty)
 all:
 	awk 'FNR==1 && NR!=1{print ""}1' $(patsubst %,%.mk,$(BUILD_CONFIG)) \
 		| sed 's/^ERLANG_MK_VERSION = .*/ERLANG_MK_VERSION = $(ERLANG_MK_VERSION)/' \
-		| sed 's:^DEFAULT_WITHOUT = .*:DEFAULT_WITHOUT = $(WITHOUT):' > $(ERLANG_MK)
+		| sed 's:^ERLANG_MK_WITHOUT = .*:ERLANG_MK_WITHOUT = $(WITHOUT):' > $(ERLANG_MK)
 
 ifdef p
 # Remove p from the list of variables since that conflicts with bootstrapping.
--- a/Makefile
+++ b/Makefile
@@ -30,7 +30,8 @@ ERLANG_MK_VERSION = $(shell git describe --tags --dirty)
 
 all:
 	awk 'FNR==1 && NR!=1{print ""}1' $(patsubst %,%.mk,$(BUILD_CONFIG)) \
-		| sed 's/^ERLANG_MK_VERSION = .*/ERLANG_MK_VERSION = $(ERLANG_MK_VERSION)/' > $(ERLANG_MK)
+		| sed 's/^ERLANG_MK_VERSION = .*/ERLANG_MK_VERSION = $(ERLANG_MK_VERSION)/' \
+		| sed 's/^DEFAULT_WITHOUT = .*/DEFAULT_WITHOUT = $(WITHOUT)/' > $(ERLANG_MK)
 
 ifdef p
 # Remove p from the list of variables since that conflicts with bootstrapping.
--- a/core/core.mk
+++ b/core/core.mk
@@ -194,6 +194,7 @@ ERLANG_MK_BUILD_CONFIG ?= build.config
 ERLANG_MK_BUILD_DIR ?= .erlang.mk.build
 
 WITHOUT ?= $(DEFAULT_WITHOUT)
+WITHOUT := $(strip $(WITHOUT))
 
 erlang-mk:
 	git clone $(ERLANG_MK_REPO) $(ERLANG_MK_BUILD_DIR)
@@ -201,7 +202,7 @@ ifdef ERLANG_MK_COMMIT
 	cd $(ERLANG_MK_BUILD_DIR) && git checkout $(ERLANG_MK_COMMIT)
 endif
 	if [ -f $(ERLANG_MK_BUILD_CONFIG) ]; then cp $(ERLANG_MK_BUILD_CONFIG) $(ERLANG_MK_BUILD_DIR)/build.config; fi
-	$(MAKE) -C $(ERLANG_MK_BUILD_DIR) WITHOUT="$(WITHOUT)"
+	$(MAKE) -C $(ERLANG_MK_BUILD_DIR) WITHOUT='$(WITHOUT)'
 	cp $(ERLANG_MK_BUILD_DIR)/erlang.mk ./erlang.mk
 	rm -rf $(ERLANG_MK_BUILD_DIR)
 
--- a/core/core.mk
+++ b/core/core.mk
@@ -18,6 +18,9 @@ ERLANG_MK_FILENAME := $(realpath $(lastword $(MAKEFILE_LIST)))
 
 ERLANG_MK_VERSION = 1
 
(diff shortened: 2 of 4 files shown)


benoitc:
I'm not sure right now how to add a test. Any idea?

essen:
Variables in unix are whitespace separated. It should therefore be `WITHOUT="index core/index"` (or if set from inside the Makefile, without the double quotes).

About the test, don't worry about it for now. there's no tests for this part yet, and it sounds a little tricky. I need to think about it.

essen:
Docs would be good though. :-)

benoitc:
@essen thanks for the review. I updated previous patch to uses a whitespace instead of a comma to split the list of the plugin to filter. 

Also I added some documentation in 2 commits to make sure is it OK to edit both. First one add a note in the getting started documentation on how to bootstrap with the `WITHOUT`  variable. Second line do the same but for the `updating`  doc. Let me know if it's OK for you.

benoitc:
I am wondering though if using a comma wouldn't be a better UX though. Using a coma removes the need of using the quotes somehow. Thoughts?

essen:
Best UX is consistency. And we're speaking 40 years of consistency here. :-)

benoitc:
> On 04 Apr 2016, at 09:46, Loïc Hoguin <notifications@github.com> wrote:
> 
> Best UX is consistency. And we're speaking 40 years of consistency here. :-)
> 
> —
> 

"," are also used in unix world for exactly such purpose though :) (this coming from a BSD user).  It makes things simpler to write for a user. 

Anyway i've no preference for one or the another. Actually the change to use whitespace just breaks in zsh. I'm trying to understand why. Will come back to you in a mn.

essen:
Don't forget the way Makefiles work where variables are lists of words separated by whitespaces. :-)

benoitc:
@essen fixed the whitespace handling :)

Makefile
```diff
@@ -12,8 +12,12 @@
 # ACTION OF CONTRACT, NEGLIGENCE OR OTHER TORTIOUS ACTION, ARISING OUT OF
 # OR IN CONNECTION WITH THE USE OR PERFORMANCE OF THIS SOFTWARE.
 
+WITHOUT ?=
+empty := $(subst ,, )
```
essen: Actually `empty :=` is enough. :-) See https://github.com/ninenines/erlang.mk/blob/master/core/core.mk#L125

essen:
I'll get back to you with a good test later today.

benoitc:
updated version that handle the case when WITHOUT is not set... (a cafe in between). Working as expected now.

essen:
I think I will write the test directly, it'll be simpler. Can you squash the commits? Thanks.

benoitc:
@essen done :)

essen:
I think there's more to be done. Don't worry about it I'll do it, putting notes here for later.

* We need to put the WITHOUT variable in the Makefile on bootstrap
* We need to use the WITHOUT variable when upgrading Erlang.mk with `make erlang-mk`

The point is that you don't need to worry about anything, just have WITHOUT when starting and then you're set.

I'm also thinking perhaps we can get rid of the build.config file entirely. The WITHOUT variable is basically a better way to customize Erlang.mk than the build.config file. If the Erlang.mk Makefile has the info in a variable directly, then this makes the implementation simpler since we can just use filter-out and such.

@dumbbell Any thoughts about this? This would have been a much simpler way to customize RabbitMQ too I believe. Though we don't need this anymore for RabbitMQ I'll be glad to hear about your thoughts. :-)

benoitc:
right.. I forgot to commit that part . If you want to remove the build.config, maybe it worth to add a` WITH` also? so you can add your plugins?

benoitc:
also the `empty = subs(..` part is needed on some platforms :)

essen:
We have moved toward providing plugins in dependencies so you don't really need much more customization than that now.

`empty = subst(..` on what platform? `empty :=` has been used in Erlang.mk for a while now and there's not been any issue reported, and it works on all supported platforms (Linux, OSX, FreeBSD and Windows).

benoitc:
when using with the sed it wasn't enough on osx...

essen:
Ah with sed. Yeah that would explain it.

dumbbell:
Hi!

I like this `WITHOUT` feature. I agree it's more convenient than `build.config`: as an end-user, you don't need to maintain a copy of the default `build.config` and keep it up-to-date. This applies to RabbitMQ.

To me, this solution is a good enough alternative to #331.

dumbbell:
I forgot to say: I tested it on FreeBSD and it works.

essen:
Thanks for the feedback!

benoitc:
@essen also re-tested on mac and ubuntu using my branch. There is no need to add the WITHOUT variable on the erlang.mk file, it will be passed to the Makefile. I can take cara bout adding to the makefile generated using the bootstrap if you want.

```shell
[test] wget https://raw.githubusercontent.com/ninenines/erlang.mk/master/erlang.mk                                                                                                                                            11:36:19
--2016-04-07 11:36:25--  https://raw.githubusercontent.com/ninenines/erlang.mk/master/erlang.mk
Resolving raw.githubusercontent.com... 103.245.222.133, 23.235.37.133, 23.235.43.133, ...
Connecting to raw.githubusercontent.com|103.245.222.133|:443... connected.
HTTP request sent, awaiting response... 200 OK
Length: 444 [text/plain]
Saving to: 'erlang.mk'

erlang.mk                                                 100%[=====================================================================================================================================>]     444  --.-KB/s    in 0s

2016-04-07 11:36:25 (38.5 MB/s) - 'erlang.mk' saved [444/444]

[test] make -f erlang.mk WITHOUT=index                                                                                                                                                                                        11:37:45
git clone https://github.com/benoitc/erlang.mk .erlang.mk.build
Cloning into '.erlang.mk.build'...
remote: Counting objects: 5875, done.
remote: Compressing objects: 100% (49/49), done.
remote: Total 5875 (delta 7), reused 0 (delta 0), pack-reused 5822
Receiving objects: 100% (5875/5875), 2.86 MiB | 1.41 MiB/s, done.
Resolving deltas: 100% (3566/3566), done.
Checking connectivity... done.
Checking out files: 100% (554/554), done.
cd .erlang.mk.build && git checkout -b fix/537 --track origin/fix/537
Branch fix/537 set up to track remote branch fix/537 from origin.
Switched to a new branch 'fix/537'
if [ -f build.config ]; then cp build.config .erlang.mk.build; fi
cd .erlang.mk.build && /Applications/Xcode.app/Contents/Developer/usr/bin/make
awk 'FNR==1 && NR!=1{print ""}1' core/core.mk core/index.mk core/deps.mk plugins/erlydtl.mk plugins/protobuffs.mk core/erlc.mk core/docs.mk core/rel.mk core/test.mk core/compat.mk plugins/asciidoc.mk plugins/bootstrap.mk plugins/c_src.mk plugins/ci.mk plugins/ct.mk plugins/dialyzer.mk plugins/edoc.mk plugins/escript.mk plugins/eunit.mk plugins/relx.mk plugins/shell.mk plugins/triq.mk plugins/xref.mk plugins/cover.mk \
                | sed 's/^ERLANG_MK_VERSION = .*/ERLANG_MK_VERSION = 2.0.0-pre.2-113-gfa7b2c2/' > erlang.mk
cp .erlang.mk.build/erlang.mk ./erlang.mk
```

essen:
There's no need because when bootstrapping it's passed on, but that is not the case when you have WITHOUT defined in the Makefile and are upgrading Erlang.mk with `make erlang-mk`.

essen:
Yes feel free to do the work, I'm probably not going to finish/merge before this week-end. Thanks!

benoitc:
@essen done ^^

core/core.mk
```diff
@@ -18,6 +18,9 @@ ERLANG_MK_FILENAME := $(realpath $(lastword $(MAKEFILE_LIST)))
 
 ERLANG_MK_VERSION = 1
 
+# Plugins ignored
```
dumbbell: "Ignored plugins" is better if my english is correct.

doc/src/guide/updating.asciidoc
```diff
@@ -61,3 +61,10 @@ You can also name the file differently or put it in a separate folder
 by modifying the value for `ERLANG_MK_BUILD_CONFIG`. You can also
 tell Erlang.mk to use a different temporary directory by changing
 the `ERLANG_MK_BUILD_DIR` variable.
+
+Alternatively you can filter a plugin in the build config by using the
```
dumbbell: "filter a plugin out"?

This applies to the sentence two lines below as well.

dumbbell:
About the documentation, should the `WITHOUT` variable become the preferred way to disable Erlang.mk components? In other words, should `WITHOUT` be documented first, then `build.config` as the alternate method?

Makefile
```diff
@@ -22,7 +30,8 @@ ERLANG_MK_VERSION = $(shell git describe --tags --dirty)
 
 all:
 	awk 'FNR==1 && NR!=1{print ""}1' $(patsubst %,%.mk,$(BUILD_CONFIG)) \
-		| sed 's/^ERLANG_MK_VERSION = .*/ERLANG_MK_VERSION = $(ERLANG_MK_VERSION)/' > $(ERLANG_MK)
+		| sed 's/^ERLANG_MK_VERSION = .*/ERLANG_MK_VERSION = $(ERLANG_MK_VERSION)/' \
+		| sed 's/^DEFAULT_WITHOUT = .*/DEFAULT_WITHOUT = $(WITHOUT)/' > $(ERLANG_MK)
```
dumbbell: You can use a single sed call instead of two, by using the `-e` flag.

Also, this doesn't work if you run `make WITHOUT=plugins/edoc` because `/` is the regex separator. I suggest you use another separator such as `|`.

With both comments:
```shell
sed \
  -e 's|^ERLANG_MK_VERSION = .*|ERLANG_MK_VERSION = $(ERLANG_MK_VERSION)|' \
  -e 's|^DEFAULT_WITHOUT = .*|DEFAULT_WITHOUT = $(WITHOUT)|' \
  > $(ERLANG_MK)
```

core/core.mk
```diff
@@ -18,6 +18,9 @@ ERLANG_MK_FILENAME := $(realpath $(lastword $(MAKEFILE_LIST)))
...
+DEFAULT_WITHOUT = ""
```
dumbbell: I would call this variable `ERLANG_MK_WITHOUT` because it's an internal detail, used during `make erlang-mk`.

benoitc:
added changes from @dumbbell in latest commits. Thanks for the suggestions

essen:
I think build.config should be deprecated for a while and then removed.

dumbbell:
Thanks @benoitc! It works fine here.

emqplus:
Really like this feature, Thank you @benoitc :)

benoitc:
what is the status of this PR? Anything I should do?

essen:
The PR is fine I think, I just got busy around the time you submitted it and then it got forgotten. I'll put it back to the top of the list.

essen:
I've merged this with a few tweaks and added tests for it. Thanks!

PR #549, opened 2016-04-17
Add tag extraction of packages and make search using tags

crownedgrouse:
`make packages`  now create test/tags.index file with a compact reverse packages index of erlang-mk tags on all packages tested.
(tested with `make -i packages` as far some are failing)

`make search` can now use  tag="..." to filter out packages matching tag(s) before use of q="..." optionally.


crownedgrouse:
to be done :  download tags.list  in local .erlang.mk  directory but need your thoughts on this.
simple wget on github or file created on make erlang-mk ?

crownedgrouse:
ping ?

essen:
Still nothing. :-) I'm almost done with Cowboy 2.0.0-pre.4 and will do a full look at all pending tickets in a couple weeks.

crownedgrouse:
Nice to ear this :>)

crownedgrouse:
Hi Loïc,
Any planning on the currents PRs for erlang.mk ? Is there some necessary work to be done before, like some prioritary bugs to fix ? A populated https://github.com/ninenines/erlang.mk/milestones might be great to follow you developpement directions .

essen:
You seem to have written commands inside the test Makefile, and not have corresponding tests for the feature being added.

crownedgrouse:
this is not intend to be used by users, but only by you at you end to create the package index to be used by users. So I'm not sure how to add a test for this.

crownedgrouse:
btw if your CI is right, I have to update the PR due to conflicts.

essen:
If it's meant to be used by me, it should be in Makefile and not test/Makefile. And it should have a test so I don't push an invalid file to unsuspecting users. I suppose a search test finding things using tags would be a way.

crownedgrouse:
Ok sorry, you are talking on the search options added. 
To add tests  on search options, I need to know how the package index will be distributed to users, in .erlang.mk probably, but at erlang.mk bootstrap ? or being updated on user demand with a curl command ?
For the package index creation, I will think how to add a test, but this might imply to launch the package index creation (a bit long !) and this is maybe not what you want ...

crownedgrouse:
I precise that it is in test/Makefile because it have to be applied on all packages, the "same" way you are testing them on compilation

essen:
I don't know. :-) I guess we should split the package index (properly this time) first before doing this.

crownedgrouse:
I mean I take the opportunity of the packages compilation test to apply the tag extraction at the end . If I have to split, I will duplicate code to download the package and compile and search tags. Not smart at all MHO.

essen:
I don't mean you splitting, I mean the files in index/* should be delivered separately and not part of the erlang.mk file.

crownedgrouse:
ok, currently the file (for users 's tag search usage) is expected in .erlang.mk/tags.index  at user side. So I need to download it from erlang.mk repo . this imply to create a special update or use the make erlang-mk command. your thought ?

crownedgrouse:
do you need somebody work on the index package extraction separated for erlang.mk body ? I'm not  sure if there is an issue or PR on this.  As far I remember index files was not inside the erlang.mk body in the past. do you have a specification ?

essen:
I think there's tickets, and yes someone needs to work on a specification first. I was looking at how OPAM does things (and are going to do things with regard to signing packages etc.)

crownedgrouse:
I think I could help on implementation, but I think you , as the owner of this project,  are the most legitimate to write  the big lines of the specification.  OPAM look nice, but package signature imply an architecture using PGP probably or worst X509 chains. A big thing to maintain.

essen:
It is but it's not like Erlang has an alternative to begin with. And even if, say, my PGP key is the only one signing anything at the beginning, it's still important to get the design right.

crownedgrouse:
Yes the design is more important for the moment.
Coming back on the subject of the PR.  
The .erlang.mk/tags.index is a single file index for tags on packages, if is 'independant' on how package index is handled .  
For now the file is created under test/ after running 'make -i packages'.
Should I move this file elsewhere, maybe under index/ to be consistent with its usage ? Then this file could be downloaded at same time than index files (whatever the solution chosen for index). 
this is a little need to sign this file , in all case.

essen:
Put it in a new directory. I was thinking of putting kerl, relx and others in their own directory too, and they could be fetched directly from the Erlang.mk repo too (and benefit from commit signing).

crownedgrouse:
ok ,  is 'tags/' ok for you  under root directory?
 or do you want something like misc/tags  misc/kerl  misc/relx etc...

essen:
Something like misc, but with a better name.

crownedgrouse:
maybe i did not understood. you mean on user side  under  .erlang.mk or did you though on the source file on your repo .

essen:
In the repo.

crownedgrouse:
opt/tags/  ?

crownedgrouse:
Are you OK with my last post (opt/ directory) ?
Or do you think something else ?

essen:
We need a directory to put all the external data and tools, not just tags but also relx, kerl and so on. This way it can just be `cp -r` in the appropriate place.

crownedgrouse:
Yes you already told me this in the thread. I just need to know the name of this root directory so that I can make a sub directory tags/ under it.
I thought to opt/ but can be ext/ or a better name if you have one.

essen:
Dunno. :-) Waiting for suggestions. Do you really need a subdirectory? Isn't it just one file?

crownedgrouse:
I'm lost.  I can move the tags file under the main root directory xxx/. I thought you wanted separated subdirectories for tags, kerl,  relx. Note some futur external things might need a subdirectory because coming with several files.
I can suggest things, but what comes to me right now is:  givingup/...

essen:
Sorry I'm not more forward, I'm still pretty busy and will be in the foreseeable future. Anyway a directory name is not important, what's important is the things I mentioned early on about you adding things in test/Makefile instead of Makefile, and the lack of test. The rest can be taken care of later.

crownedgrouse:
ok, but there is currently nothing in Makefile, only in core/index.mk and core/core.mk for the runtime (the search of packages from tags) by users, and already in test/Makefile for 'server' side to create the tags.index file to put at disposal for the runtime on the repo.
I will try to add a test. I will put the tags file index in a rsc/  (resource) directory. The name can to changed after.

essen:
"already in test/Makefile for 'server' side to create the tags.index file to put at disposal for the runtime on the repo" - yes, that's what I'm saying, it should be in Makefile, not test/Makefile. I understand that it seems convenient that it's there, but it's really awkward because then it'd run every time I run tests on every single environment.

PR #566, opened 2016-06-08
Add PROJECT_MOD var to make the app callback module configurable

jflatow:
I have many projects in which the application callback module (mod key in .app) is not called $(PROJECT)_app.erl, this is a very small change to make it configurable.


essen:
Thanks. This needs a test also (test/core_app.mk).

jflatow:
OK, I've added a test.

essen:
Merged, thanks!

PR #568, opened 2016-06-09
Fix several problems with the erlydtl plugin:

jflatow:
DTL_SUFFIX:
 - make sure it is actually used
 - add a test for it (and one combined with other options)

DTL_PATH:
 - handle correctly with or without trailing /
 - don't hard-code the `doc_root` opt - the erlydtl default is better

DTL_FILES:
 - the user should be able to generate template files before compiling
   therefore do not define any targets for these, so that they may do so

### DIFF
--- a/plugins/erlydtl.mk
+++ b/plugins/erlydtl.mk
@@ -15,19 +15,14 @@ dtl_verbose = $(dtl_verbose_$(V))
 
 # Core targets.
 
-DTL_FILES = $(sort $(call core_find,$(DTL_PATH),*.dtl))
+DTL_PATH := $(abspath $(DTL_PATH))
+DTL_FILES = $(call core_find,$(DTL_PATH),*.dtl)
 
 ifneq ($(DTL_FILES),)
 
-ifdef DTL_FULL_PATH
-BEAM_FILES += $(addprefix ebin/,$(patsubst %.dtl,%_dtl.beam,$(subst /,_,$(DTL_FILES:$(DTL_PATH)%=%))))
-else
-BEAM_FILES += $(addprefix ebin/,$(patsubst %.dtl,%_dtl.beam,$(notdir $(DTL_FILES))))
-endif
-
-# Rebuild templates when the Makefile changes.
-$(DTL_FILES): $(MAKEFILE_LIST)
-	@touch $@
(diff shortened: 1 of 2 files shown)


essen:
Thanks!

About doc_root, can you explain what the change does exactly?

Otherwise looks good and assuming tests pass it'll be merged quickly. Just want to understand that point. :-)

jflatow:
Sure - it was hardcoded to "templates", which presumably was meant to be DTL_PATH. According to https://github.com/erlydtl/erlydtl, the default is the compiled template's directory, which I thought made sense. Although I wonder if this is right now that you mention it!

essen:
Before making that kind of change I'd like to have tests with more complex templates that may be affected, for example those using the extend or include directives.

jflatow:
Yea that makes sense. I'll add tests for it.

essen:
Thanks!

jflatow:
I added a test, and also manually confirmed that it works as expected - includes are relative to the original file.

essen:
Thanks, I'll take a closer look tomorrow.

essen:
Sorry it took so long. I'm adding this to run the tests now.

essen:
Failures are unrelated to your changes (except on Windows, but Windows fails with or without the changes). It's late so I will do the merge tomorrow.

essen:
OK I have an issue with this: "the user should be able to generate template files before compiling therefore do not define any targets for these, so that they may do so"

This shouldn't be one or the other, we need to recompile everything when the Makefile changes. It doesn't have a test in the erlydtl suite so you probably didn't realize you were breaking it. For example, if we change DTL_OPTS, we want to recompile the files automatically.

I will merge without the removal of lines 28-30 and we can work in a new PR that includes a test for generating templates in the Makefile (your use case) and a test for the current rebuild behavior.

The solution in erlc.mk has been to use double colon `::` to allow both. The same is probably possible in this case too.

essen:
Merged, minus the change mentioned above. Thanks!

jflatow:
OK thanks for merging and maintaining in general :) Not sure about the handling of DTL_FILES, but regardless makes sense to do it separately.

essen:
I'm busy catching up with tickets so feel free to try with changing the colon to a double colon, adding tests and see if it works for your code generation case. Probably won't have a look at it for a while otherwise.

PR #572, opened 2016-06-27
merge PortSpec env into each file output env

icehess:
We should append environment vars(LDFLAGS for example) to defined environment in for each output file.

for example LDFLAGS is set a few lines above for OS X, but if output defined LDFLAGS it will be replaced.

### DIFF
--- a/core/deps.mk
+++ b/core/deps.mk
@@ -374,7 +374,7 @@ define dep_autopatch_rebar.erl
 					"%.o: %.C\n\t$$\(CXX) -c -o $$\@ $$\< $$\(CXXFLAGS) $$\(ERL_CFLAGS) $$\(DRV_CFLAGS) $$\(EXE_CFLAGS)\n\n",
 					"%.o: %.cc\n\t$$\(CXX) -c -o $$\@ $$\< $$\(CXXFLAGS) $$\(ERL_CFLAGS) $$\(DRV_CFLAGS) $$\(EXE_CFLAGS)\n\n",
 					"%.o: %.cpp\n\t$$\(CXX) -c -o $$\@ $$\< $$\(CXXFLAGS) $$\(ERL_CFLAGS) $$\(DRV_CFLAGS) $$\(EXE_CFLAGS)\n\n",
-					[[Output, ": ", K, " = ", ShellToMk(V), "\n"] || {K, V} <- lists:reverse(MergeEnv(FilterEnv(Env)))],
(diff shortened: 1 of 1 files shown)


essen:
Sounds OK in theory.

icehess:
@essen any update on this?

essen:
The problem with this PR is that I need to run the build of all packages to make sure we don't break anything. I've setup CI again for Erlang.mk and will even enable nightly builds tonight; assuming everything goes right this PR will run tomorrow.

Sorry it took so long to set things up. I can't really do this on my laptops (especially during summer), the package test suite takes 2 fairly intensive hours at least. Will keep you informed.

icehess:
Thanks for your update, yes running the test is my issue as well :)

I have this issue with compiling exml as a dependency for Kazoo on Mac OS X. exml sets LDFLAGS and CFLAGS in its rebar.conf, and they overwrite erlang.mk generated Makefile counterpart vars which in Mac OS X case needs to have `-flat_namespace -undefined suppress` included in LDFLAGS.

I'm thinking for another approach. Right now erlang.mk writes those vars for each target, but we can define those vars at beginning and then for each target just appending their vars if they have any.

essen:
I'm sure Erlang.mk could use less brute force tactics when it comes to dealing with rebar projects. :-)

essen:
Well nightly build is setup. If all goes well it starts in 8 minutes. Otherwise I'll start it manually. Either way it'll run tonight, and your PR tomorrow. Cheers!

essen:
Alright! Here are the results:

On Arch Linux, FreeBSD and OSX, the following packages break: `brod` and `kafka_protocol` (compile errors).

On FreeBSD, somehow this fixes `restc`.

On OSX this fixes `exml`.

I do not have a report for Windows at this point, I think I misconfigured the build. Doesn't really matter though, I do not expect much changes from this PR.

Full reports are at https://builds.ninenines.eu/logs/erlang-dot-mk-packages/ with build 11 the nightly and 20 your PR.

I will investigate the new errors. You may try them on your end with `make check p=brod V=3` for example.

essen:
`brod` fails because it depends on `kafka_protocol`, which reduces to one project having issues.

essen:
The issue is unrelated and was fixed by adding a missing package to the index. I will merge this, thanks!

essen:
Merged, thanks!

icehess:
oh great! thanks.

PR #585, opened 2016-07-30
Add package erlangzmq to the index

drozzy:
Resubmitting PR for erlangzmq with dialyzer warnings fixed.


drozzy:
Sorry, but I'm having a hard time understanding the CI output. Could you let me know what's not working?

I'm not sure if I mentioned, but this uses Erlang/OTP 19.

essen:
Don't worry about CI. All I care about is that 'make check p=erlangzmq' works. If it does, then this can be merged.

drozzy:
Oh ok. Did it work?
Just out of curiosity what does that check?

essen:
It checks that the package compiles and loads properly. And it did. Merged, thanks!

Issue #587, opened 2016-08-05
Env in .app


nevar:
In .app.src we can define env (for default value, or some other use-case).
For now .app.src is legacy and all app define got to Makefile.
What about env?

essen:
There's been some discussions about it. The question is what form should this take. A variable in the Makefile? A separate file?

essen:
Perhaps we can find a solution that makes it easy to generate a "default" sys.config file afterwards.

crownedgrouse:
.app.src.script was an interesting approach by rebar. This could make more compliant both tools ?

essen:
.app.src is pretty bad, .app.src.script is even worse. All the .script things are very error prone, hard to debug and require too much maintenance whenever rebar changes again.

To be honest a variable in the Makefile is probably enough, considering Make 4 and above allow you to do something like `var = $(file cowboy.env)` should you want a separate file.

nevar:
Make PR for PROJECT_ENV?

essen:
PROJECT_ENV sure, additional keys though which ones do you refer to? The ones that are left are so rarely used I'm not sure it's worth adding until someone complains.

nevar:
PROJECT_$(KEY) = $(VALUE) =>
```erl
{application, name, [ {$(KEY), $(VALUE)}]}
```

essen:
That makes no sense. If you're going to want this you can just use a .app file.

The only keys not configurable right now, excluding env, are maxP, maxT, included_applications, start_phases and runtime_dependencies. The first four are exceedingly rare, and the last one I've never seen outside OTP's applications. The only thing we miss is env.

nevar:
okay. Then I will make PR for PROJECT_ENV.

dumbbell:
Hi!

@nevar: I played a bit with your patches (#588 + #589) and it works well. Thank you!

I was wondering, should the scope of this issue be wider to allow users to put any keys in the `.app` file? Or should other keys (like the less used ones or the non-standard ones used in RabbitMQ master branch) be handled by another variable, eg. `PROJECT_APP_EXTRA`? I don't have a strong opinion on either way.

essen:
Can you link to some variables RabbitMQ uses?

dumbbell:
Sure, here is an example, taken from [rabbitmq-shovel](https://github.com/rabbitmq/rabbitmq-shovel/blob/master/src/rabbitmq_shovel.app.src#L13):
```erlang
[...]

Issue #366, opened 2015-09-18
env.src file for generation of ebin/.app file


lehoff:
The env part of a .app file is not really suited for translation into a Makefile syntax so the suggestion is to create a src/env.src file that holds the definition of the env key for the .app file.

A real-life example:
``` erlang
 {env, [
         %% Cluster name
         {cluster_name, "default"},

         %% Default location for ring, cluster and other data files
         {platform_data_dir, "data"},

         %% Default ring creation size.  Make sure it is a power of 2,
         %% e.g. 16, 32, 64, 128, 256, 512 etc
         {ring_creation_size, 64},

         %% Default gossip interval (milliseconds)
         {gossip_interval, 60000},

         %% Target N value
         {target_n_val, 4},

         %% Default claims functions
         {wants_claim_fun, {riak_core_claim, default_wants_claim}},
         {choose_claim_fun, {riak_core_claim, default_choose_claim}},

         %% Vnode inactivity timeout (how often to check if fallback vnodes
         %% should return their data) in ms.
         {vnode_inactivity_timeout, 60000},

         %% Number of VNodes allowed to do handoff concurrently.
         {handoff_concurrency, 2},

         %% Disable Nagle on HTTP sockets
         {disable_http_nagle, true},

         %% Handoff IP/port
         {handoff_port, 8099},
         {handoff_ip, "0.0.0.0"},

         %% Disterl buffer sizes in bytes.
         %% These sizes (3*128*1024 & 6*128*1024) were
         %% derived from a limited amount of testing in a
         %% 10GE environment, and may need tuning for your
         %% network and workload. In particular they're likely
         %% too small to be optimal for larger object sizes.
         {dist_send_buf_size, 393216},
         {dist_recv_buf_size, 786432},

     %% Exometer defaults
     {exometer_defaults,
      [
       {['_'], histogram, [{options,
                [{histogram_module, exometer_slot_slide},
                 {keep_high, 500}]}
                ]}
      ]}.
```

essen:
Yep, file should probably be this, minus the first line.

lehoff:
yeah, just the list would probably be a better choice

essen:
Having that kind of file would also make it easy to generate a global sys.config for all deps.

essen:
[...]

PR #588, opened 2016-08-05
Add PROJECT_ENV

nevar:
See #587

### DIFF
--- a/core/erlc.mk
+++ b/core/erlc.mk
@@ -67,7 +67,8 @@ define app_file
 	{id$(comma)$(space)"$(1)"}$(comma))
 	{modules, [$(call comma_list,$(2))]},
 	{registered, []},
-	{applications, [$(call comma_list,kernel stdlib $(OTP_DEPS) $(LOCAL_DEPS) $(DEPS))]}
+	{applications, [$(call comma_list,kernel stdlib $(OTP_DEPS) $(LOCAL_DEPS) $(DEPS))]},
+	{env, $(subst \,\\,$(PROJECT_ENV))}
 ]}.
 endef
 else
@@ -79,7 +80,8 @@ define app_file
 	{modules, [$(call comma_list,$(2))]},
 	{registered, [$(call comma_list,$(PROJECT)_sup $(PROJECT_REGISTERED))]},
 	{applications, [$(call comma_list,kernel stdlib $(OTP_DEPS) $(LOCAL_DEPS) $(DEPS))]},
-	{mod, {$(PROJECT_MOD), []}}
+	{mod, {$(PROJECT_MOD), []}},
+	{env, $(subst \,\\,$(PROJECT_ENV))}
 ]}.
 endef
 endif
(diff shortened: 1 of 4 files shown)


core/erlc.mk
```diff
@@ -60,14 +60,16 @@ app:: clean deps $(PROJECT).d
 endif
 
 ifeq ($(wildcard src/$(PROJECT_MOD).erl),)
+PROJECT_ENV ?= []
```
essen: Define this here instead: https://github.com/nevar/erlang.mk/blob/38fe63dafdd8e9c902e5b242da4dd02ab0f0067f/core/core.mk#L27

test/core_app.mk
```diff
@@ -28,6 +28,27 @@ core-app-appsrc-change: build clean
 	$t rm $(APP)/EXPECT
 endif
 
+ifndef LEGACY
+core-app-env: build clean
```
essen: Alphabetical order.

test/core_app.mk
```diff
@@ -1,6 +1,6 @@
 # Core: Building applications.
 
-CORE_APP_CASES = appsrc-change asn1 auto-git-id erlc-exclude erlc-opts erlc-opts-filter error generate-erl generate-erl-include generate-erl-prepend hrl hrl-recursive makefile-change mib no-app no-makedep project-mod pt pt-erlc-opts xrl xrl-include yrl yrl-include
+CORE_APP_CASES = appsrc-change asn1 auto-git-id erlc-exclude erlc-opts erlc-opts-filter error generate-erl generate-erl-include generate-erl-prepend hrl hrl-recursive makefile-change mib no-app no-makedep project-mod pt pt-erlc-opts xrl xrl-include yrl yrl-include env
```
essen: Alphabetical order.

essen:
Small tidbits but otherwise looks fine. Also missing a section in the user guide.

Thanks!

nevar:
updated

nevar:
Faced with issue:
1. printf for app file don't escape % (will send PR for that)
2. For old make file ``$(shell cat file.env)`` eat all \n and if file has erlang comment then .app not right. ``$(file < file.env)`` look fine (don't eat \n). May by it is good to note that in document and add example how read file.env. I end up with ``$(shell cat file.env | sed 's/$$/\\n/')``

nevar:
Also it will be good if we can autoadd ``file.env`` as dependency of ``ebin/$(PROJECT).app``

essen:
What's the thing about % and " about? Not sure I get it.

nevar:
If I have in file.env: ``[{test, "%"}]``, then printf will insert parameter for %. Same apply for ``PROJECT_DESCRIPTION = test %``.  Need make ``%`` -> ``%%``. This can fixed for all app_file template in https://github.com/ninenines/erlang.mk/blob/master/core/erlc.mk#L244. And this is separate from this PR I think.

If I have in file.env: ``[{test, "\"test\""}]``, then after insert to app template printf command will be: ``printf "[..skip..]{env, [{test, \"\\"test\\"\"}]}[..skip]" and after process we will get: ``{env, [{test, "\test\"}]``. Same apply for ``PROJECT_DESCRIPTION = test "\"test\""``. Need ``\`` -> ``\\``. Not sure we wan't fix that for PROJECT_DESCRIPTION.

essen:
Right, thanks.

nevar:
Change printf "..." to printf '...' for app_file.
Add escape / for PROJECT_ENV

nevar:
#366

nevar:
For get env from file and make version < 4: ``PROJECT_ENV = $(subst \n,$(newline),$(shell cat FILE.env | sed 's/$$/\\n/'))``

essen:
CI currently running on this.

essen:
Merged, thanks!!

PR #589, opened 2016-08-08
Escape % in app_file template

nevar:
See https://github.com/ninenines/erlang.mk/pull/588#issuecomment-238157332


essen:
CI will run this also.

essen:
Merged, thanks!!

Issue #238, opened 2015-05-11
Put all temporary files and tools in a .erlang.mk directory


essen:
Instead of putting everything intermediate in ., put them hidden in .erlang.mk.

This includes the package index file, third party tools (relx, xrefr...) and possibly also temporary Erlang modules required for some operations.

essen:
This includes the PLT file.

essen:
Tagging as bug because it's damn annoying.

essen:
The bulk of the work is done, closing. Thanks!

PR #593, opened 2016-08-24
Hide generated file by erlang.mk

nevar:
#238

### DIFF
--- a/plugins/c_src.mk
+++ b/plugins/c_src.mk
@@ -6,7 +6,7 @@
 # Configuration.
 
 C_SRC_DIR ?= $(CURDIR)/c_src
-C_SRC_ENV ?= $(C_SRC_DIR)/env.mk
+C_SRC_ENV ?= $(LOCAL_ERLANG_MK_TMP)/env.mk
 C_SRC_OUTPUT ?= $(CURDIR)/priv/$(PROJECT)
 C_SRC_TYPE ?= shared
 
@@ -124,6 +124,7 @@ endif
 
 ifneq ($(wildcard $(C_SRC_DIR)),)
 $(C_SRC_ENV):
+	$(verbose) mkdir -p $(dir $(C_SRC_ENV))
 	$(verbose) $(ERL) -eval "file:write_file(\"$(call core_native_path,$(C_SRC_ENV))\", \
 		io_lib:format( \
 			\"ERTS_INCLUDE_DIR ?= ~s/erts-~s/include/~n\" \
--- a/plugins/ci.mk
+++ b/plugins/ci.mk
@@ -3,7 +3,7 @@
 
 .PHONY: ci ci-setup distclean-kerl
 
-KERL ?= $(CURDIR)/kerl
+KERL ?= $(ERLANG_MK_TMP)/kerl
 export KERL
 
 KERL_URL ?= https://raw.githubusercontent.com/yrashk/kerl/master/kerl
--- a/plugins/cover.mk
+++ b/plugins/cover.mk
@@ -12,7 +12,7 @@ COVER_MODS = $(notdir $(basename $(call core_ls,ebin/*.beam)))
 
 test-build:: $(LOCAL_ERLANG_MK_TMP)/ct.cover.spec
 
-$(LOCAL_ERLANG_MK_TMP)/ct.cover.spec:
+$(LOCAL_ERLANG_MK_TMP)/ct.cover.spec: | $(LOCAL_ERLANG_MK_TMP)
 	$(verbose) echo Cover mods: $(COVER_MODS)
 	$(gen_verbose) printf "%s\n" \
 		'{incl_mods,[$(subst $(space),$(comma),$(COVER_MODS))]}.' \
@@ -58,7 +58,7 @@ coverdata-clean:
 	$(gen_verbose) rm -f *.coverdata ct.cover.spec
 
 # Merge all coverdata files into one.
-$(LOCAL_ERLANG_MK_TMP)/all.coverdata: $(COVERDATA)
+$(LOCAL_ERLANG_MK_TMP)/all.coverdata: $(COVERDATA) | $(LOCAL_ERLANG_MK_TMP)
 	$(gen_verbose) $(ERL) -eval ' \
 		$(foreach f,$(COVERDATA),cover:import("$(f)") == ok orelse halt(1),) \
 		cover:export("$(LOCAL_ERLANG_MK_TMP)/$@"), halt(0).'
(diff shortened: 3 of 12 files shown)


nevar:
Not tested for executable file: kerl, xrefr

nevar:
May be *.o from c_src also hide?

essen:
Why?

nevar:
Don't like add c_src/*.o to .gitignore

essen:
I wouldn't be against an OBJDIR variable, but they're not really temporary files. Intermediate files may be used in more than one way. I notice a few other things in your PR and will comment inline.

plugins/c_src.mk
```diff
@@ -6,7 +6,7 @@
 # Configuration.
 
 C_SRC_DIR ?= $(CURDIR)/c_src
-C_SRC_ENV ?= $(C_SRC_DIR)/env.mk
+C_SRC_ENV ?= $(LOCAL_ERLANG_MK_TMP)/env.mk
```
essen: This file may be committed, it's not temporary. Really depends on the project.

plugins/ci.mk
```diff
@@ -3,7 +3,7 @@
 
 .PHONY: ci ci-setup distclean-kerl
 
-KERL ?= $(CURDIR)/kerl
+KERL ?= $(ERLANG_MK_TMP)/kerl
```
essen: This one is correct, but... does this work and fetch kerl properly?

plugins/cover.mk
```diff
@@ -1,7 +1,7 @@
 # Copyright 2015, Viktor Söderqvist <viktor@zuiderkwast.se>
 # This file is part of erlang.mk and subject to the terms of the ISC License.
 
-COVER_REPORT_DIR = cover
+COVER_REPORT_DIR = $(LOCAL_ERLANG_MK_TMP)/cover
```
essen: The cover directory is an output, not a temporary file. It's similar to the logs directory from Common Test. There's no point in hiding it.

plugins/cover.mk
```diff
@@ -10,15 +10,15 @@ ifdef CT_RUN
 # All modules in 'ebin'
 COVER_MODS = $(notdir $(basename $(call core_ls,ebin/*.beam)))
 
-test-build:: $(TEST_DIR)/ct.cover.spec
+test-build:: $(LOCAL_ERLANG_MK_TMP)/ct.cover.spec
```
essen: ct.cover.spec may be committed and customized.

plugins/cover.mk
```diff
@@ -51,17 +51,17 @@ help::
 
 # Plugin specific targets
 
-COVERDATA = $(filter-out all.coverdata,$(wildcard *.coverdata))
+COVERDATA = $(filter-out $(LOCAL_ERLANG_MK_TMP)/all.coverdata, $(wildcard $(LOCAL_ERLANG_MK_TMP)/*.coverdata))
```
essen: coverdata files are output and may be used afterwards by various tools. It's a bit awkward to have them in the root directory though, would probably be better in cover/

plugins/ct.mk
```diff
@@ -11,6 +11,7 @@ ifneq ($(wildcard $(TEST_DIR)),)
 else
 	CT_SUITES ?=
 endif
+CT_LOGDIR ?= $(LOCAL_ERLANG_MK_TMP)/ct
```
essen: The fact that rebar3 hides the logs/ directory in _build/test/ is my biggest daily annoyance with it. I really don't want to change that. Making it configurable sure, but the default is great.

plugins/ct.mk
```diff
@@ -33,13 +34,12 @@ CT_RUN = ct_run \
...
 	-dir $(TEST_DIR) \
-	-logdir $(CURDIR)/logs
+	-logdir $(CT_LOGDIR)
 
 ifeq ($(CT_SUITES),)
 ct: $(if $(IS_APP),,apps-ct)
 else
-ct: test-build $(if $(IS_APP),,apps-ct)
-	$(verbose) mkdir -p $(CURDIR)/logs/
+ct: test-build $(if $(IS_APP),,apps-ct) | $(CT_LOGDIR)
```
essen: Best to avoid unnecessary changes/make separate commits.

plugins/dialyzer.mk
```diff
@@ -5,7 +5,7 @@
 
 # Configuration.
 
-DIALYZER_PLT ?= $(CURDIR)/.$(PROJECT).plt
+DIALYZER_PLT ?= $(LOCAL_ERLANG_MK_TMP)/.$(PROJECT).plt
```
essen: I'm not convinced the PLT is temporary. It could very well be committed to speed up dialyzing on large projects.

plugins/relx.mk
```diff
@@ -5,7 +5,7 @@
 
 # Configuration.
 
-RELX ?= $(CURDIR)/relx
+RELX ?= $(ERLANG_MK_TMP)/relx
```
essen: Same as kerl.

plugins/xref.mk
```diff
@@ -11,7 +11,7 @@ else
 	XREF_ARGS := -c $(XREF_CONFIG)
 endif
 
-XREFR ?= $(CURDIR)/xrefr
+XREFR ?= $(ERLANG_MK_TMP)/xrefr
```
essen: Same as kerl.

essen:
I have numerous unmerged changes to my local copy. Some are very similar to this, some could benefit from your changes. I will try to do a good mix of both tomorrow.

nevar:
I think we need two configurable variable:
1. ERLANG_MK_TMP for temporary file: xrefr, relx, kerl
2. _WHAT_EVER_ for output: .$(PROJECT).plt, ct, *.coverdata, ct.cover.spec (by default it is $CURDIR)

essen:
So far kerl and relx have been moved to $ERLANG_MK_TMP. xrefr hasn't; I am waiting to take a closer look/document it, it seems pretty basic and perhaps we can do more than that with xref.

Other than that, this remains:

* Move the coverdata files to the cover/ directory
* Review/document cover (and perhaps do something about ct.cover.spec, perhaps not)
* Allow configuring CT logs location

Feel free to submit a separate PR for the CT logs; you already have the work done there, all that's needed is a test and renaming the variable to CT_LOGS_DIR to be more in line with the rest of the variables.

I will be doing the cover stuff soon.

essen:
We should move `ebin/test` into `.erlang.mk/test` also.

essen:
I'll close this. There's most likely a few bits and pieces to improve still, but the bulk of the work was done and we can take each remaining issue independently, if necessary. Thanks!

PR #594, opened 2016-08-29
dialyzer.mk: Use the shell to parse command line args

dumbbell:
Splitting arguments on `-` was dangerous: if a path contained such a character, it would be split and the second half thrown away by the filtering.

Now, `$(ERLC_OPTS)` is passed to erl(1) after `-extra`. Thus arguments parsing is done by the shell and we only have to call `init:get_plain_arguments/0` to get them as a list.


essen:
Merged, thanks!

PR #595, opened 2016-09-02
Add support for customized prologue when compiling yecc (yrl) files.

lordnull:
The documentation for erlc allows the use of `-I` to define a custom prologue to use when compiling yacc source files. This is an option available in rebar. However, there does not exist a way to define a custom prologue in erlang.mk. This adds a variable one can set allowing one to not only define a custom prologue, buy any additional erlc options one my want.


essen:
Wow thanks! And sorry for the late reply, I missed a few things on Github due to email issues. I will run the tests now.

essen:
Tests look good. Failures are unrelated.

lordnull:
I didn't think they were related, but good to know for sure. Thanks for the review :).

essen:
Merged, thanks!

PR #596, opened 2016-09-03
Add package chumak to the index

drozzy:
This is the same package as `erlangzmq` only name and license are changed.

If possible, please delete `erlangzmq` from the index.

### DIFF
--- a/index/chumak.mk
+++ b/index/chumak.mk
@@ -1,0 +1,7 @@
+PACKAGES += chumak
+pkg_chumak_name = chumak
+pkg_chumak_description = Pure Erlang implementation of ZeroMQ Message Transport Protocol.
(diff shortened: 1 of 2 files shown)


essen:
Please add a commit doing "git rm index/erlangzmq.mk" and that'll do it. :-)

drozzy:
Do you mean in the commit message?
Like: git commit -m "git rm index/erlangzmq.mk"

essen:
I mean add a separate commit that deletes erlangzmq using the command I provided.

drozzy:
Got it :-)

essen:
Merged with a change, 1.1.1 didn't exist so I changed the version to master.

drozzy:
Sorry, my bad. Forgot to push the tags up. Master is ok :+1:

PR #597, opened 2016-09-08
plugins/xref.mk: Fix typo in variable `XREFR_ARGS`

dumbbell:


essen:
Merged because obvious, thanks!

PR #608, opened 2016-10-07
Support C compiler flag overrides for dependencies

fhunleth:
This change makes it possible to override the ERL_CFLAGS and ERL_LDFLAGS
Makefile variables in project dependencies. For the main project, the
ability to override variables associated with C compilation already
existed in plugins/c_src.mk. With this change, the C compiler flag
semantics are similar for both the main project and dependencies.

This is important for cases where the automatically determined values
for the ERTS include and library directories are incorrect. This happens
in cross-compiled environments where they point to the host's include
and library directories. However, the desired behaviour is to have them
point to those that have been cross-compiled for the target.

### DIFF
--- a/core/deps.mk
+++ b/core/deps.mk
@@ -334,9 +334,9 @@ define dep_autopatch_rebar.erl
 		[] -> ok;
 		_ ->
 			Write("\npre-app::\n\t$$\(MAKE) -f c_src/Makefile.erlang.mk\n"),
-			PortSpecWrite(io_lib:format("ERL_CFLAGS = -finline-functions -Wall -fPIC -I \\"~s/erts-~s/include\\" -I \\"~s\\"\n",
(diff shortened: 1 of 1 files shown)


essen:
Sounds good, running tests to make sure but I don't foresee any problem.

I would love to hear more about your cross compiling experience. A step by step would be great for example. This is something I would really like to have in the user guide.

fhunleth:
Thanks - the change seems benign, but I'm a novice use of erlang.mk and I'm getting into it for a project that I'm helping out with. I'll keep that in mind about contributing back some documentation. That definitely would have been useful for me to read, but I do have to say that my experience with erlang.mk has been very positive so far. It meshes well with the other build tools I use since it uses `make`.

essen:
Looking forward to it.

Also the tests seemed fine (regardless of all the failures you can see, that are unrelated), so this will get merged quickly.

essen:
Merged, thanks!

PR #612, opened 2016-10-17
Fix erlexec application name

acammack-r7:
erlexec now requires that particular name in order to successfully boot.


essen:
Thanks!

PR #616, opened 2016-10-20
Support PROJECT_APP_EXTRA_KEYS to add keys to the .app file

dumbbell:
This allows to add standard keys which do not have a corresponding `PROJECT_` Makefile variable, as well as non-standard keys.

### DIFF
--- a/test/core_app.mk
+++ b/test/core_app.mk
@@ -1,6 +1,6 @@
 # Core: Building applications.
 
-CORE_APP_CASES = appsrc-change asn1 auto-git-id env erlc-exclude erlc-opts erlc-opts-filter error generate-erl generate-erl-include generate-erl-prepend hrl hrl-recursive makefile-change mib name-special-char no-app no-makedep project-mod pt pt-erlc-opts xrl xrl-include yrl yrl-header yrl-include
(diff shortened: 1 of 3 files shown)


test/core_app.mk
```diff
@@ -229,6 +229,45 @@ core-app-env: build clean
...
+	$t $(MAKE) -C $(APP) $v
+
+	$i "Check that the application was compiled correctly"
+	$t $(ERL) -pa $(APP)/ebin/ -eval " \
+		ok = application:load($(APP)), \
+		AppFile = filename:join(code:lib_dir($(APP), ebin), atom_to_list($(APP)) ++ \".app\"), \
+		{ok, [App]} = file:consult(AppFile), \
+		{application, $(APP), Props} = App, \
+		'\\\$$my_app' = proplists:get_value(non_standard_atom, Props),\
+		\"\\\"test_\\tvalue\\\"\" = proplists:get_value(non_standard_string, Props),\
```
essen: You don't need to use file:consult etc. This is what application:get_key/1,2 is for.

dumbbell: `application:get_key/{1,2}` only looks for standard keys. It doesn't return non-standard ones.

essen: Wow. Learn something new every day. Fair enough.

essen:
Tests are good. Preparing travel so no merge just yet.

essen:
Merged, thanks!

PR #617, opened 2016-10-24
installation.asciidoc: Fix Erlang package name on FreeBSD

dumbbell:


essen:
Well spotted. Looks like my notes were incorrect.

essen:
Merged, thanks!

PR #634, opened 2016-11-23
Makefile: Set LC_COLLATE=C when building erlang.mk

dumbbell:
When the index files are concatenated, (`index/*` in `build.config`), their order is related to the collation locale used by the shell. By explicitely setting this to the C locale, we make sure that the order is
consistent, no matter the user's locale.

It is especially useful in teams where users and systems have different locales.


essen:
Sounds good. Merge tomorrow if tests show no issues.

essen:
Merged, thanks!

PR #642, opened 2016-12-14
plugins/escript.mk: Remove $(ESCRIPT_FILE) in distclean-escript

dumbbell:
... not `$(ESCRIPT_NAME)`, which can be set to another value.


essen:
Should probably improve a test with a check that it gets removed before merging.

dumbbell:
Ok, I will add one.

dumbbell:
Done!

essen:
Tests are good.

essen:
Merged, to be pushed in a few. Thanks!

PR #643, opened 2016-12-14
core/erlc.mk: Quote application name in .app

dumbbell:


essen:
Also good.

essen:
And merged. Thanks!

PR #655, opened 2017-01-05
core/erlc.mk: added src subfolders support in makedep.erl

okebkal:


okebkal:
done in test/core_makedep.mk

PR #657, opened 2017-01-11
core/erlc.mk: added src subfolders support in makedep.erl

okebkal:


essen:
Hello, the tests you added don't seem to work, except on OSX apparently. Do you have access to a Linux or BSD?

okebkal:
I did the test under Debian Linux. Could you provide more detailed information about the fails?

essen:
I think the failure is because you use echo with \n in it. In the rest of the test suite we use printf to make sure it works everywhere. Try changing this, I'll re-run the tests and see if that fixes it.

okebkal:
Ok, change done, please re-run the test.

okebkal:
I see already some fails here, but could you give me the log details?

essen:
Probably unrelated, I'll get back to you when I get more time.

essen:
Merged locally, thanks!

Issue #658, opened 2017-01-11
Passing asn.1 compiler options?


klajo:
Hi,

I'm looking for a way of passing ASN.1 compiler options (there are options for choosing ASN.1 binary encoding and the way values are returned form the encoder/decoder for example).

Is this the way to do it (see below) or is there another, better way?  I can submit a pull request if this is the way to go.

/Klas

```make
diff --git a/core/erlc.mk b/core/erlc.mk
index 51255b1..65ef71b 100644
--- a/core/erlc.mk
+++ b/core/erlc.mk
@@ -12,6 +12,8 @@ COMPILE_FIRST_PATHS = $(addprefix src/,$(addsuffix .erl,$(COMPILE_FIRST)))
 ERLC_EXCLUDE ?=
 ERLC_EXCLUDE_PATHS = $(addprefix src/,$(addsuffix .erl,$(ERLC_EXCLUDE)))
 
+ERLC_ASN1_OPTS ?=
+
 ERLC_MIB_OPTS ?=
 COMPILE_MIB_FIRST ?=
 COMPILE_MIB_FIRST_PATHS = $(addprefix mibs/,$(addsuffix .mib,$(COMPILE_MIB_FIRST)))
@@ -104,7 +106,7 @@ ERL_FILES += $(addprefix src/,$(patsubst %.asn1,%.erl,$(notdir $(ASN1_FILES))))
 
 define compile_asn1
 	$(verbose) mkdir -p include/
-	$(asn1_verbose) erlc -v -I include/ -o asn1/ +noobj $(1)
+	$(asn1_verbose) erlc -v -I include/ -o asn1/ +noobj $(ERLC_ASN1_OPTS) $(1)
 	$(verbose) mv asn1/*.erl src/
 	$(verbose) mv asn1/*.hrl include/
 	$(verbose) mv asn1/*.asn1db include/
```

essen:
Sounds good, yes. Just needs a test to confirm it works and won't get broken in the future and it can be merged quickly.

essen:
Closing!

PR #659, opened 2017-01-12
Set ASN.1 compilation options using ERLC_ASN1_OPTS

klajo:
`ERLC_ASN1_OPTS` can be used to pass compiler options when compiling
ASN.1 files. By default, Erlang.mk will leave this empty, but it can
be redefined in a Makefile.

This aims to implement the support requested in #658.


essen:
OK merging too.

essen:
Merged, thanks!

klajo:
Thanks!

PR #663, opened 2017-01-23
Fix Archlinux Installation Instructions

### LINKED ISSUE #662: [DOCS] Archlinux install
at 1.1.1. Linux you suggest `$ pacman -Sy erlang git make`.
But using `-Sy` is [a terrible idea](https://bbs.archlinux.org/viewtopic.php?id=89328).

ingwinlu:
Suggesting -Sy is a terrible idea[1]. This commit replaces it with -Syu
which is the default on how packages should be added on Archlinux.

Resolves #662.

[1]: https://bbs.archlinux.org/viewtopic.php?id=89328

### DIFF
--- a/doc/src/guide/installation.asciidoc
+++ b/doc/src/guide/installation.asciidoc
@@ -22,7 +22,7 @@ The commands to install packages vary between distributions:
 
 .Arch Linux
 [source,bash]
-$ pacman -Sy erlang git make
+$ pacman -S erlang git make
 
 ==== FreeBSD
 


essen:
Are you sure we want to say -Syu though? I was expecting -S. I'm not sure telling users to update everything is that great for just getting started.

ingwinlu:
Hi,

I assumed you don't want a whole intro into Archlinux package management in
the installation section so I put -Syu as it is the only thing guaranteed
to work all the time. Partial upgrades can be very bad (-Sy). (-Su) might
fail because the version of your local repository is not the one on the
servers.

You need to resync your local repository(y), (u)pdate everything because of
the whole rolling release thingy and then install the new packages (-S).

essen:
Fair enough. I'll merge when I get the time. Thanks!

ingwinlu:
Update to use `-S` instead of `-Syu` . I agree with you that this is the way forward. (Also the one used in archlinux wiki for example).

essen:
I thought this was merged!

essen:
Thanks!

Issue #667, opened 2017-02-08
Improve the way Erlang.mk is autopatched


essen:
Currently we replace the erlang.mk file with another containing only `include ../../erlang.mk` (the path is not hardcoded, but gives this by default).

This is not ideal. It would be better if we didn't have to change anything to make it work. There are two possible solutions.

One is through the `MAKEFILES` variable which allows to load Makefiles unconditionally. But this would load them in separate places from the include, can be source of errors and would require some support in the deps' erlang.mk itself.

The other is via the `-I` option, which instructs Make where the .mk files are located. If we temporarily rename erlang.mk to erlang.mk.bak, then run our command on the dep, then rename erlang.mk.bak to erlang.mk again, then we can do everything without any potentially problematic changes to the dependency.

/cc @ingwinlu

ingwinlu:
[Described behaviour](https://www.gnu.org/software/make/manual/html_node/Include.html#Include) in the gnu make docs.

essen:
Alternatively we can replace the `include erlang.mk` line to be something like `include $(if $(ERLANG_MK),$(ERLANG_MK),erlang.mk)` and export `ERLANG_MK` to point to the currently running Erlang.mk file. This doesn't allow us to do without any change to the project, but does have the advantage of being a neutral change (one you can commit safely). Thoughts?

essen:
My issue with `-I` and other options is that they're not neutral. If the dep compilation fails, and we moved erlang.mk to erlang.mk.bak, we have no way of putting it back (not easily/hacky at least).

essen:
It works pretty well. I think if we modify the tutorial to have this line directly (though erlang.mk would of course still work) and have the line swap done otherwise, we'll have a much cleaner solution since it'll select the correct Erlang.mk file depending on context and will not modify any file for any new users.

essen:
Done in https://github.com/ninenines/erlang.mk/commit/3f8688e0d0526b74993d64a166e9ba8559c687e4

Thanks!

PR #668, opened 2017-02-08
[Unfinished] Add symlinking dependencies

ingwinlu:
Implements symlinking dependencies.

Note that currently autopatch replaces the symlinked erlang.mk incorrectly, resulting in a failure of the included test if  `NO_AUTOPATCH_ERLANG_MK` is not set.

Ref #667


essen:
Removed the no autopatch, squashed and merged. Thanks!

Issue #558, opened 2016-05-16
How to specify the build order of multiple apps in one project?


emqplus:
The project's structure:

```
apps/
    core_app
    extend_app1
    extend_app2
```

The extend_app1 and extend_app2 depend on core_app. How to ensure the core_app is built first? Thanks.

essen:
Did you try adding LOCAL_DEPS=core_app to both extend applications? I'm not sure this works yet.

emqplus:
@essen, Thanks for your quick reply. The LOCAL_DEPS cannot work, which is only used to generate the .app file:

```
define app_file
{application, $(PROJECT), [
        {description, "$(PROJECT_DESCRIPTION)"},
        {vsn, "$(PROJECT_VERSION)"},$(if $(IS_DEP),
        {id$(comma)$(space)"$(1)"}$(comma))
        {modules, [$(call comma_list,$(2))]},
        {registered, []},
        {applications, [$(call comma_list,kernel stdlib $(OTP_DEPS) $(LOCAL_DEPS) $(DEPS))]}
]}.
endef 
```

essen:
Right it's not been implemented yet. That's how it will eventually work though so do put them there.

If it's a blocker I can try to get it fixed in the afternoon.

emqplus:
Hi @essen, I add a 'LOCAL_DEPS_DIRS' variable in erlang.mk to work around this issue. Please check the patch:

```
diff --git a/erlang.mk b/erlang.mk
index 01da469..42bb8f1 100644
--- a/erlang.mk
+++ b/erlang.mk
@@ -4090,7 +4090,8 @@ dep_repo = $(patsubst git://github.com/%,https://github.com/%, \
        $(if $(dep_$(1)),$(word 2,$(dep_$(1))),$(pkg_$(1)_repo)))
 dep_commit = $(if $(dep_$(1)_commit),$(dep_$(1)_commit),$(if $(dep_$(1)),$(word 3,$(dep_$(1))),$(pkg_$(1)_commit)))

-ALL_APPS_DIRS = $(if $(wildcard $(APPS_DIR)/),$(filter-out $(APPS_DIR),$(shell find $(APPS_DIR) -maxdepth 1 -type d)))
+LOCAL_DEPS_DIRS = $(foreach a,$(LOCAL_DEPS),$(if $(wildcard $(APPS_DIR)/$(a)),$(APPS_DIR)/$(a)))
[...]

PR #673, opened 2017-02-17
compile apps in the right order by looking at their LOCAL_DEPS

geogriff:
how does this look as a fix for #558?

### DIFF
--- a/core/deps.mk
+++ b/core/deps.mk
@@ -26,6 +26,7 @@ dep_repo = $(patsubst git://github.com/%,https://github.com/%, \
 	$(if $(dep_$(1)),$(word 2,$(dep_$(1))),$(pkg_$(1)_repo)))
 dep_commit = $(if $(dep_$(1)_commit),$(dep_$(1)_commit),$(if $(dep_$(1)),$(word 3,$(dep_$(1))),$(pkg_$(1)_commit)))
 
+LOCAL_DEPS_DIRS = $(foreach a,$(LOCAL_DEPS),$(if $(wildcard $(APPS_DIR)/$(a)),$(APPS_DIR)/$(a)))
 ALL_APPS_DIRS = $(if $(wildcard $(APPS_DIR)/),$(filter-out $(APPS_DIR),$(shell find $(APPS_DIR) -maxdepth 1 -type d)))
 ALL_DEPS_DIRS = $(addprefix $(DEPS_DIR)/,$(foreach dep,$(filter-out $(IGNORE_DEPS),$(BUILD_DEPS) $(DEPS)),$(call dep_name,$(dep))))
 
@@ -48,9 +49,6 @@ dep_verbose = $(dep_verbose_$(V))
 
 # Core targets.
 
-ifdef IS_APP
-apps::
-else
 apps:: $(ALL_APPS_DIRS)
 ifeq ($(IS_APP)$(IS_DEP),)
 	$(verbose) rm -f $(ERLANG_MK_TMP)/apps.log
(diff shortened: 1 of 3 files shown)


essen:
Nice! Sounds good. Just needs a few tests covering the different cases you speak about in the comments. Thanks in advance!

geogriff:
1) looks like the test core-deps-apps covers LOCAL_DEPS containing an app at the toplevel. i'll add one to cover (a) a single app-to-app dependency, make sure both are built, (b) circular app-to-app dependency, make sure make succeeds and both are built, and (c) toplevel LOCAL_DEPS missing, and check all apps are built.

2) question: do we really want to change the semantics of LOCAL_DEPS in the way that i did, where not all local apps are necessarily compiled if at least one is listed in the toplevel LOCAL_DEPS? this might break existing projects with strange layouts or without proper LOCAL_DEPS declarations..

essen:
1- I think we need more than "both are built". I think your tests should include a parse_transform in the local dep, and it being used in the other dep, so that it fails if the compile order is incorrect. And just to be on the safe side, have two tests with two of those apps named (for example) app1 and app2, and have the tests differ in that app1 has the parse transform in one test, and app2 in the other test. This way we'll know it's not just because the app names were properly ordered by chance when using wildcard.

2- The default behavior should probably be to first compile in order, and then compile everything else.

geogriff:
those tests look good?

essen:
Will get back to you on Tuesday. Please ping the ticket on Tuesday just in case. Thanks!

geogriff:
ping

ingwinlu:
also, this change is awesome. gj @geogriff !

geogriff:
done.

my use case, in case you are wondering, is that the code for my cluster is split up into different apps, and each type of server has a different set of apps, but i prefer to keep all the code for my cluster in one repository. plus they're not generic enough to be useful to anyone outside of my cluster. so they're all in the apps/ folder, but only some of them get built and included in different releases. this fix really helps compile time (only needed apps are built) and then obviously was necessary to get compilation order right.

ingwinlu:
One thing this pr does not tackle is -includes.
Building on the example in #558:
* core_app defines a behaviour
* extend_app uses this behaviour

Resulting in the behaviour not being found.

Additionally all the 'subapps' are only compilable from the top level (minor).

geogriff:
good points. 1) the most common cases of this are going to be covered by LOCAL_DEPS; just include extend_app in the LOCAL_DEPS of core_app, because when are you going to want to use a behaviour from another app but not include it as a runtime dependency? the OTP way of thinking about behaviours, anyway, seems to be to use the facilities of the behaviour module (e.g. gen_server and friends) or _at least_ the facilities of another module in the app (e.g. cowboy_handler). we can add LOCAL_BUILD_DEPS for this, and for parse transforms which is a tad more valid of a use case, but I'd argue that this just complicates erlang.mk more and it's a solution looking for a problem.

I propose these two issues be issued by another pull request when the need arises, and we note the issue in the docs if you feel that is necessary?

essen:
I'm not sure what this issue mentioned is. If you need a behavior from another application, you *have* to depend on it using DEPS or LOCAL_DEPS. Behaviors are more than just compile time checks.

ingwinlu:
I messed up. Behaviours work as expected with this PR.

After a quick test to see if behaviours work in an [example repo](https://github.com/ingwinlu/erlang.mk/commit/eb01732237f15f65c8e919b32b3011d1a64e1dc3) I saw that it works flawlessly. Further investigation showed that the issue was with the repo where i tried it.

essen:
I will not look in more details until next week so I'll let @ingwinlu or someone else do further review/experiments in the meantime. Make sure you squash the commits to help me merge more easily.

Thanks!

geogriff:
@ingwinlu are you saying that the issue was not with this pr?

will i see either of you at erlang factory 2017? ;-)

essen:
If you mean SF, no, too far. Did it a few times in the past but it's not worth it for me.

geogriff:
@essen ah, i figured. this year looks great (joe armstrong, robert virding, carl hewitt, larry wall!) but just watching the videos is probably sufficient :-P

essen:
I think this year looks very boring. Not much I can apply to my work in there.

geogriff:
haha, so pragmatic...

geogriff:
back to the PR, is there anything else I need to do to get this merged once-and-for-all?

ingwinlu:
Well there has already been a request for a squash of the commits.

The rest is ok from my side.

essen:
Merged locally! Thanks.

PR #674, opened 2017-03-05
fix removing deps.log at toplevel when in a multi-app layout

geogriff:
`rm deps.log` was not being executed, because of 1) the ifeq around
it, checking for an empty ALL_DEPS_DIRS, which i believe was a logical
error; and 2) deps depended on apps, which means apps were compiled
before its recipe was executed, and the `rm deps.log` would be
executed after it had been written to by recursive make of the
apps.

the important reason to remove deps.log is so that dependencies get
remade when they change. otherwise you'll get mysterious errors about
missing dependencies if they need to be rebuilt, or worse, have wrong
stale beam files bundled in your release.

tests core-deps-apps and core-deps-apps-only now manually clean the
cowlib dep and rebuild at the top-level to make sure cowlib gets
rebuilt. both tests indeed fail without this fix.

this attempts to fix #1 by removing the ifeq, and #2 by having deps
and apps depend on deps.log, with deps.log 'built' (but actually
removed) by a double-colon rule with no prerequisites (which means its
recipe always be run).

### DIFF
--- a/core/deps.mk
+++ b/core/deps.mk
@@ -1,7 +1,7 @@
 # Copyright (c) 2013-2016, Loïc Hoguin <essen@ninenines.eu>
 # This file is part of erlang.mk and subject to the terms of the ISC License.
 
-.PHONY: distclean-deps
+.PHONY: distclean-deps clean-tmp-deps.log
 
 # Configuration.
 
@@ -51,7 +51,7 @@ dep_verbose = $(dep_verbose_$(V))
 ifdef IS_APP
 apps::
 else
-apps:: $(ALL_APPS_DIRS) $(ERLANG_MK_TMP)/deps.log
+apps:: $(ALL_APPS_DIRS) clean-tmp-deps.log
 ifeq ($(IS_APP)$(IS_DEP),)
 	$(verbose) rm -f $(ERLANG_MK_TMP)/apps.log
 endif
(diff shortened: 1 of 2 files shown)


geogriff:
not entirely sure this is the best fix, but please let me know what you think. this issue has bit me quite a few times, and i've had to be very careful to work around it every time in production.

essen:
@dumbbell Thoughts?

core/deps.mk
```diff
@@ -72,16 +72,14 @@ endif
...
 endif
 
-ifneq ($(SKIP_DEPS),)
-deps::
-else
-ifeq ($(ALL_DEPS_DIRS),)
-deps:: apps
-else
-deps:: $(ALL_DEPS_DIRS) apps
+$(ERLANG_MK_TMP)/deps.log::
```
dumbbell: Because this recipe doesn't create the target file, I would use another name for the target, such as `remove-tmp-deps.log` and make it `.PHONY` (in which case you shouldn't need the double-colon).

core/deps.mk
```diff
@@ -72,16 +72,14 @@ endif
...
 ifeq ($(IS_APP)$(IS_DEP),)
 	$(verbose) rm -f $(ERLANG_MK_TMP)/deps.log
 endif
+ifneq ($(SKIP_DEPS),)
```
dumbbell: I would insert a blank line above to clearly separate recipes.

core/deps.mk
geogriff: agreed

essen:
Merged locally, thanks!

PR #676, opened 2017-03-15
Make eunit in multi-apps project should exit non-zero if any test is failed

headcrabmeat:


essen:
Thanks. It's possible I won't look too thoroughly for a couple weeks, I managed to free up some time starting next week only.

headcrabmeat:
@essen , sure, no problem

essen:
Merged locally with a few tweaks in the tests (you can just do `$t ! command` to ensure it returns an error :-). Closing, thanks!

PR #677, opened 2017-03-17
Update relx up to v3.23.0

headcrabmeat:
Some relx release contain `relx` escript some of them do not.
I think common approach would be to whether to clone the repository or whether to get packed sources and build `relx` with `rebar3`.

### DIFF
--- a/plugins/relx.mk
+++ b/plugins/relx.mk
@@ -8,7 +8,7 @@
 RELX ?= $(ERLANG_MK_TMP)/relx
 RELX_CONFIG ?= $(CURDIR)/relx.config
 
-RELX_URL ?= https://github.com/erlware/relx/releases/download/v3.19.0/relx
+RELX_URL ?= https://github.com/erlware/relx/releases/download/v3.23.0/relx
 RELX_OPTS ?=
 RELX_OUTPUT_DIR ?= _rel
 RELX_REL_EXT ?=
--- a/test/plugin_relx.mk
+++ b/test/plugin_relx.mk
@@ -114,28 +114,57 @@ else
 	$t test `$(APP)/tmp/bin/$(APP)_release rpcterms test test` = old
 endif
 
-	$i "Move the relup tarball to the release directory"
+	$i "Check that it's 1 avalible version"
+ifeq ($(PLATFORM),msys2)
+	$t test `$(APP)/tmp/bin/$(APP)_release.cmd versions |wc -l` = "2"
+else
+	$t test `$(APP)/tmp/bin/$(APP)_release versions |wc -l` = "2"
+endif
+
+	$i "Copy the relup tarball to the release directory"
 	$t mkdir $(APP)/tmp/releases/2
-	$t mv $(APP)/_rel/$(APP)_release/$(APP)_release-2.tar.gz $(APP)/tmp/releases/2/$(APP)_release.tar.gz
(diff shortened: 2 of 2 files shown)


essen:
Erlang.mk can build relx, don't need rebar3.

That said, I was thinking of having a solution that doesn't make us too dependent on upstreams. There's a few options, the most sensible probably to add relx to BUILD_DEPS (`BUILD_DEPS += relx`) when the relx.config file exists.

Another option is to build and bundle it ourselves, but I'm not looking forward to doing that. The automatic approach is probably better (and could allow users to define a specific version or repository of Relx they want, even if it's not a tag).

essen:
I tried with v3.22.3 (which has a relx file) but `make check c=relx-relup V=3` fails here. Can you try?

essen:
Problem disappeared, I think my environment was just fucked up. I'll update it now.

essen:
Nevermind I was running the wrong test.

essen:
3.22.0 also has the issue. Seems there's a dependency on `compiler` (really?) but even if I add it the test fails. Someone please investigate, otherwise I'm fine with leaving the current version for the time being.

headcrabmeat:
Hey, sorry for a delay. I asked guy who releases relx to add binaries. 
So I think there is no need in this PR.

essen:
Yeah but that doesn't solve the fact that most recent Relx broke release upgrades, at least for us.

headcrabmeat:
Do you mean `relup` & `reldown`, don't you? Could you please describe a little bit precisely.

essen:
Try "make check c=relx-relup V=3" after changing the relx version to v3.22.3 and you'll see what I mean.

headcrabmeat:
I did, It failed. Seems that `relup` tests are actually outdated. Need more time to investigate. It's strange that CI shows it's "ok"

essen:
It says OK because this is just the noop that succeeded, I haven't enabled the tests on this PR.

headcrabmeat:
I think I found out the root cause. I'm updating testcases. Will share it soon.

BTW, I checked your approach with `DEPS += relx` it works.

headcrabmeat:
I talked with the guy who releases `relx`. He confirmed that relx binary could be added to the release page of relx project. So what's the best approach?

essen:
Just fetching the binary for now should do.

essen:
I see you added compiler to the relx.config template. That's mad. Why would release upgrades require the compiler? Release upgrades (without relx) only require sasl. Relx should be fixed to not require it instead.

essen:
Sent a patch to relx. Will wait for it to be merged before merging this PR. https://github.com/erlware/relx/pull/589

headcrabmeat:
Ops. That's my bad. I compared it with a `relx.config` in my project. It had compiler. 
So I thought like "Ok, new rules..." and added it to the bootstrap.

essen:
You are not the one mad here. The madness was probably accidental on their part too. The PR has been merged there so it's just a matter of waiting for the next version now.

essen:
Relx 3.23 has just been released, can you update your PR with it? Shouldn't need compiler anymore.

headcrabmeat:
Sure, no problem

headcrabmeat:
It's done + rebased.

essen:
Merged locally, will push soon. Thanks!

headcrabmeat:
cool, thanks

Issue #678, opened 2017-03-20
Is it possible to avoid generating the tar ball?


peffis:
I'm sorry if this is more a *stupid feature request*, but I'm wondering if I have missed something. erlang.mk did not use to create a tar-ball in the past (and perhaps I've been running with old erlang.mk versions for long in my project) but now it seems to always call relx with the tar option, like so: 

```
relx-rel: $(RELX) rel-deps app
        $(verbose) $(RELX) -c $(RELX_CONFIG) $(RELX_OPTS) release tar
```

Is it possible to customize erlang.mk somehow so that I can decide when the tar ball is generated? Removing the "tar" from the line above seems to do the trick but is there some other way to do it? I have not much need for tar-balls most of the time so it is just a waste of time in the build process for me.

essen:
It should probably have an `$(if $(SFX), tar)` instead of just `tar`. Experiment and send a patch!

PR #679, opened 2017-03-21
making tar-ball generation optional with relx

peffis:
Make relx tar-ball generation optional with RELX_DISABLE_TAR environment variable as per discussion in #678. 

The old behaviour of generating the tar file is maintained is default if you do not set RELX_DISABLE_TAR. 

Usage: 
```
$ make -f erlang.mk bootstrap bootstrap-rel
$ RELX_DISABLE_TAR=1 make
```

### DIFF
--- a/plugins/relx.mk
+++ b/plugins/relx.mk
@@ -11,6 +11,11 @@ RELX_CONFIG ?= $(CURDIR)/relx.config
 RELX_URL ?= https://github.com/erlware/relx/releases/download/v3.19.0/relx
 RELX_OPTS ?=
 RELX_OUTPUT_DIR ?= _rel
+RELX_TAR ?= 1
+
+ifdef SFX
+	RELX_TAR = 1
+endif
 
 ifeq ($(firstword $(RELX_OPTS)),-o)
 	RELX_OUTPUT_DIR = $(word 2,$(RELX_OPTS))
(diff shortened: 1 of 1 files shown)


essen:
Should probably not enable the tar by default. Also I mentioned the SFX variable because when you have SFX=1 the tar is required, so your new variable conflicts with it. `make SFX=1 RELX_DISABLE_TAR=1` will break.

peffis:
Ah, I did not realize SFX actually was an existing variable that had a meaning. I thought you were just referring to some generic variable name in a pseudo-code kind of manner. I'll update the pull request. 

And since you don't worry about being backwards compatible in respect to the generation of the tar, I will make tar-ball-generation disabled by default.

essen:
It's not documented that it does and was only added when I added sfx so it should be OK.

essen:
Thanks. Tests running!

peffis:
I cannot see the details of those failing tests above unfortunately.

essen:
I'll check when they are all completed.

essen:
Ah my bad. It does say in the documentation that the tar file is always created, and there's a test for that that fails in your PR. I had completely forgotten about it. Sorry about that.

Then I suppose the way to go is to add a variable `RELX_TAR ?= 1` which is set to 1 by default and to do the test like `if ($(RELX_TAR)$(SFX),tar)`, meaning you can disable it with `make RELX_TAR=`.

Or better yet you can use ifeq/else/endif and then we can have a better `make RELX_TAR=0` to disable it, at the cost of a little more code (but that's probably better).

peffis:
Ok, no worries, I'll try this other approach

essen:
By the way, `make check c=sfx` and `make check c=relx` will test most things that may fail because of your changes.

peffis:
```
$ make check c=sfx
/Applications/Xcode.app/Contents/Developer/usr/bin/make -C test sfx
make[1]: *** No rule to make target `sfx'.  Stop.
make: *** [check] Error 2
```

...but the relx tests passed.

essen:
I guess I didn't add sfx tests yet. If relx tests are good then it should be OK. Running tests again.

Just one nitpick: remove the space after the comma on the two ifeq lines. Shouldn't change the test output, but will be more consistent with the rest of the code.

essen:
Quick glance at the tests show success. I'll continue tomorrow and merge if all is good. Thanks!

peffis:
👍

essen:
Merged locally! Thanks!

PR #687, opened 2017-04-19
Move external plugin loading at the end of erlang.mk

essen:
This allows all plugins to append to existing variables
and makes it easier to augment the functionality of the
core plugins.

This lets me write a plugin like this, for example:

``` makefile
THIS := $(dir $(realpath $(lastword $(MAKEFILE_LIST))))

SOURCES += $(THIS)nif_helpers.c
$(C_SRC_OUTPUT_FILE): $(THIS)nif_helpers.o

ifeq ($(PLATFORM),msys2)
	CFLAGS += -I"$(THIS)/compat/"
endif

CFLAGS += -I"$(THIS)"
```

Feedback welcome!

### DIFF
--- a/build.config
+++ b/build.config
@@ -40,5 +40,8 @@ plugins/xref
 plugins/cover
 plugins/sfx
 
+# External plugins.
+core/plugins
+
 # Core modules which can use variables from plugins.
 core/deps-tools
--- a/core/deps.mk
+++ b/core/deps.mk
@@ -591,18 +591,3 @@ ERLANG_MK_RECURSIVE_DOC_DEPS_LIST = $(ERLANG_MK_TMP)/recursive-doc-deps-list.log
 ERLANG_MK_RECURSIVE_REL_DEPS_LIST = $(ERLANG_MK_TMP)/recursive-rel-deps-list.log
 ERLANG_MK_RECURSIVE_TEST_DEPS_LIST = $(ERLANG_MK_TMP)/recursive-test-deps-list.log
 ERLANG_MK_RECURSIVE_SHELL_DEPS_LIST = $(ERLANG_MK_TMP)/recursive-shell-deps-list.log
-
-# External plugins.
-
-DEP_PLUGINS ?=
-
-define core_dep_plugin
--include $(DEPS_DIR)/$(1)
-
-$(DEPS_DIR)/$(1): $(DEPS_DIR)/$(2) ;
-endef
-
-$(foreach p,$(DEP_PLUGINS),\
-	$(eval $(if $(findstring /,$p),\
-		$(call core_dep_plugin,$p,$(firstword $(subst /, ,$p))),\
-		$(call core_dep_plugin,$p/plugins.mk,$p))))
--- a/core/plugins.mk
+++ b/core/plugins.mk
@@ -1,0 +1,17 @@
+# Copyright (c) 2013-2017, Loïc Hoguin <essen@ninenines.eu>
+# This file is part of erlang.mk and subject to the terms of the ISC License.
+
+# External plugins.
+
+DEP_PLUGINS ?=
+
+define core_dep_plugin
+-include $(DEPS_DIR)/$(1)
+
+$(DEPS_DIR)/$(1): $(DEPS_DIR)/$(2) ;
+endef
+
+$(foreach p,$(DEP_PLUGINS),\
+	$(eval $(if $(findstring /,$p),\
+		$(call core_dep_plugin,$p,$(firstword $(subst /, ,$p))),\
+		$(call core_dep_plugin,$p/plugins.mk,$p))))


bullno1:
Doesn't break [reload.mk](https://github.com/bullno1/reload.mk) :+1:

essen:
Merged, will push soon! Thanks for the feedback.

dumbbell:
Sorry for chiming in after the merge...

I have one problem with this change in the context of RabbitMQ: a plugin can't add dependencies anymore. For instance, we have a plugin which adds Common Test hooks and this plugin is used by all our projects (that's why we do it in a plugin instead of all the Makefiles).

What do you think of having a second kind of plugins, i.e. "early plugins", loaded near the beginning of Erlang.mk?

essen:
Can't you do DEPS += dep and then $(call dep_target,dep) or something? I think that's the only difference for this.

dumbbell:
I'll try that, thank you!

dumbbell:
I tried to add `$(call dep_target,dep)` in the plugin but it doesn't work.

First, adding `$(call dep_target,dep)` was unnecessary: appending to `$(TEST_DEPS)` was enough. Unfortunately, when the `test-deps` rule is evaluated, the value of `$(ALL_TEST_DEPS_DIRS)` only contains `$(TEST_DEPS)` defined so far, not those appended later.

I tried to add those dependencies to `test-deps` by adding the following line to the plugin:
```make
# $(ALL_PLUGIN_DEPS_DIRS) are the late dependencies' directories.
test-deps: $(ALL_PLUGIN_DEPS_DIRS)
```
But that didn't do the trick.

essen:
Which plugin is it?

dumbbell:
[`rabbitmq-build.mk`](https://github.com/rabbitmq/rabbitmq-common/blob/master/mk/rabbitmq-build.mk#L90-L102).

essen:
It seems to work for me if I do both:

``` makefile
$(foreach dep,$(TEST_DEPS),$(eval $(call dep_target,$(dep))))
test-deps: $(ALL_TEST_DEPS_DIRS)
```

Note that this ends up having some warnings about redefining targets so you'd need to restrict it to your own deps. But it otherwise seems to work fine.

dumbbell:
You're right, by repeating all targets depending the late dependencies, it appears to work:
```make
CTH_DEPS = cth_fail_fast cth_styledout
ALL_CTH_DEPS_DIRS = $(addprefix $(DEPS_DIR)/,$(CTH_DEPS))

$(foreach dep,$(CTH_DEPS),$(eval $(call dep_target,$(dep))))
ifeq ($(SKIP_DEPS),)
test-deps: $(ALL_CTH_DEPS_DIRS)
$(ERLANG_MK_RECURSIVE_TEST_DEPS_LIST): $(ALL_CTH_DEPS_DIRS)
ifneq ($(filter test,$(DEP_TYPES)),)
$(ERLANG_MK_RECURSIVE_DEPS_LIST): $(ALL_CTH_DEPS_DIRS)
endif
endif
```

The warning about duplicated targets goes away if all projects have an up-to-date copy of `erlang.mk`. But this still requires to repeat all targets depending on those added dependencies.

Also, I can't append to `$(CT_OPTS)` anymore. It works with `$(TEST_ERLC_OPTS)`, but Make recurses to build testsuites: perhaps that's why it works for this variable.

essen:
Not sure why it wouldn't work, there's nothing special done with it.

dumbbell:
Anyway, I would be more comfortable with "early plugins" which can set variables used by Erlang.mk, instead of duplicating code. Do you suspect any problem with such plugins?

essen:
Where would they be loaded? Because I think previously while they allowed using TEST_DEPS they didn't allow using some other DEPS and BUILD_DEPS.

dumbbell:
Here is the test I did:

1. I moved the `core_dep_plugin` macro just before the definition of `dep_name` in a cutting-edge `erlang.mk` and loaded "early plugins" there:

    ```make
    # ...
    REBAR_DEPS_DIR = $(DEPS_DIR)
    export REBAR_DEPS_DIR

    define core_dep_plugin
    -include $(DEPS_DIR)/$(1)

    $(DEPS_DIR)/$(1): $(DEPS_DIR)/$(2) ;
    endef

    DEP_EARLY_PLUGINS ?=

    # Load "early plugins"; defaults to a file named `early-plugins.mk`.
    $(foreach p,$(DEP_EARLY_PLUGINS),\
    	$(eval $(if $(findstring /,$p),\
    		$(call core_dep_plugin,$p,$(firstword $(subst /, ,$p))),\
    		$(call core_dep_plugin,$p/early-plugins.mk,$p))))

    dep_name = $(if $(dep_$(1)),$(1),$(if $(pkg_$(1)_name),$(pkg_$(1)_name),$(1)))
    # ...
    ```

2. I created an example project, plugin and dependency to test with it. You can find the [source files in a Gist](https://gist.github.com/dumbbell/69ccccdda854b251d47623bc80edddb7), though you'll need to adjust locations.

3. I tried the following commands:

    ```sh
    gmake distclean && gmake
    # Successfully fetched `myplugin` then `mydep`. `mydep` was built.
    # `$(ERL_OPTS)` has the expected value.

    gmake distclean && gmake mytarget
    # Successfully fetched `myplugin` and executed `mytarget`.

    gmake distclean && gmake ct
    # Successfully fetched everything and executed the testsuite with the
    # correct `$(CT_OPTS)`.
    ```

What do you think?

essen:
Sounds good.

dumbbell:
Thanks, I'll prepare a pull request then.

PR #696, opened 2017-05-10
test/core_deps.mk: Wait for node in core-deps-rel

dumbbell:
When the test check the running applications in the started release, the node may not be up yet. So loop until the command returns a list of applications (instead of "Node is not running!").

This fixes some transient failures.


essen:
Merged, push soon!

PR #697, opened 2017-05-10
core/deps.mk: Move .app in autopatch2 only if file exists

dumbbell:
This silences warnings such as:

```
mv: rename file.app to file.app.src: No such file or directory
```

They are harmless really, but still annoying.


essen:
Merged too.

PR #698, opened 2017-05-10
Make sure the build fails if a docs/rel/test/shell dep fails

dumbbell:
Before this change, the build would continue, even if a test dependency failed to build. This could lead to obscure errors in the middle of a testsuite.

With this change, the build fails immediately, exactly like when a regular dependency fails to build.

### DIFF
--- a/core/deps-tools.mk
+++ b/core/deps-tools.mk
@@ -65,22 +65,20 @@ ifeq ($(IS_APP)$(IS_DEP),)
 	$(verbose) rm -f $(ERLANG_MK_RECURSIVE_TMP_LIST)
 endif
 ifndef IS_APP
-	$(verbose) for dep in $(ALL_APPS_DIRS) ; do \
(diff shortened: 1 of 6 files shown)


essen:
I'd prefer something like what we do at https://git.ninenines.eu/erlang.mk.git/tree/core/deps.mk#n73 what do you think?

dumbbell:
I tried that first and it works. But usually, I prefer `set -e` because we don't have to worry about the commands changing. I.e., if there are commands added inside the loop, error handling will be fine.

dumbbell:
I made the same change to docs, rel and shell dependency building.

I used the same `set -e`, but I'm ok to change it to `|| exit $$?` if you prefer.

essen:
Why not the main deps? Would be good to have the same everywhere I think.

dumbbell:
In the last commit, I replaced all `|| exit $$?` with `set -e`. This should be unified now.

essen:
Merged!

PR #700, opened 2017-05-12
Support early-stage plugins through `$(DEP_EARLY_PLUGINS)`

dumbbell:
Regular plugins (`$(DEP_PLUGINS)`) are loaded near the end of Erlang.mk. This is fine when you want to modify variables initialized earlier in Erlang.mk or add new targets and variables.

However, it doesn't allow you to declare more dependencies because they are loaded too late for that.

This commit introduces a new variable, `$(DEP_EARLY_PLUGINS)`, which can be used to list plugins meant to be loaded near the beginning of Erlang.mk. Those allow to append to the list of dependencies.

They work exactly like regular plugins otherwise. The default filename loaded is `early-plugins.mk`.

See #687 for the discussion leading to this new feature.

### DIFF
--- a/core/deps.mk
+++ b/core/deps.mk
@@ -21,6 +21,22 @@ export DEPS_DIR
 REBAR_DEPS_DIR = $(DEPS_DIR)
 export REBAR_DEPS_DIR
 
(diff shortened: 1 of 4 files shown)


essen:
Merged!

PR #701, opened 2017-05-12
Support plugins local to the application

dumbbell:
If the application's Makefile specify either:
```
DEP_PLUGINS = LOCAL
```
or e.g.:
```
DEP_PLUGINS = LOCAL/mk/dist.mk
```
then load the plugin from the application instead of a dependency.

This helps when you have an application with common Erlang modules and Erlang.mk plugins: your common application can load Erlang.mk plugins exactly like other applications depending on the common application.

### DIFF
--- a/core/deps.mk
+++ b/core/deps.mk
@@ -25,9 +25,13 @@ export REBAR_DEPS_DIR
 # They both use the core_dep_plugin macro.
 
 define core_dep_plugin
+ifeq ($(2),$(PROJECT))
+-include $$(patsubst $(PROJECT)/%,%,$(1))
+else
 -include $(DEPS_DIR)/$(1)
 
 $(DEPS_DIR)/$(1): $(DEPS_DIR)/$(2) ;
+endif
 endef
 
 DEP_EARLY_PLUGINS ?=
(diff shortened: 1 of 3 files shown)


essen:
What's the difference between this and just adding the include yourself in the Makefile?

dumbbell:
Loading before/after `include erlang.mk` could have a different result than inside Erlang.mk. With this change, the user doesn't have to worry about possible changes in Erlang.mk.

Before #687, this would have been useful in rabbitmq-common because `rabbitmq-build.mk` was included differently in rabbitmq-common compared to other projects.

essen:
I'm not sure what you mean, if the location of plugin loading changes the user definitely could be hit, even with this feature. I fail to see what this solves.

dumbbell:
It ensures consistency accross projects.

essen:
But it's not consistent, you have a special notation for it. It would be a lot easier to accept if instead of LOCAL you were using `$(PROJECT)`, for example. Then it would be consistent. You also only added it to DEP_PLUGINS, it probably should be applied to both mechanisms (I know it's not merged, I'm pushing soon...)

dumbbell:
You're right, using `$(PROJECT)` is a better filtering, thanks.

If you agree with the feature, I plan to rebase it after the early plugin patch is also accepted.

dumbbell:
I also didn't write any documentation so far. I will write it if you're ok with the feature and when we agree on the implementation.

essen:
OK I pushed with the changes required for this.

dumbbell:
The commit was rebased.

essen:
Please add it to the early plugins and add docs and I'll merge it, thanks!

dumbbell:
Done!

essen:
Fixed a few typos and merged. Thanks!

PR #702, opened 2017-05-12
plugins/xref.mk: Fetch xrefr 1.1.0

dumbbell:
In particular, xrefr 1.1.0 now exits with a non-zero exit status if there are any warnings reported. This makes `make xref` fail appropriately.


essen:
Merged.

PR #706, opened 2017-06-08
Update lager repository org.

hairyhum:
[basho/lager](https://github.com/basho/lager) is unmaintained. A while ago, [erlang-lager/lager](https://github.com/erlang-lager/lager) project was started
to continue progress on lager, including support for recent OTP versions and rebar3.


essen:
Thanks!

Issue #714, opened 2017-08-09
Run command before make run and make tests


benjamin-bergia:
What would be the recommend way to run some shell commands before every make run/tests ?
I find myself starting dependencies such as database and cache instance before running or testing my project.

essen:
You can add more commands to some targets, for example:

```
tests::
    start_db

include erlang.mk
```

If it's placed before it'll run before, after/after. All targets in Erlang.mk that have the two colons `::` can be extended like this.

benjamin-bergia:
Cool exactly what I wanted. I will take a look into it

benjamin-bergia:
@essen I just noticed that the run target doesn't have two colons. I updated my erlang.mk to add a second colon and I tested with a `run::` before and after the include. Both seem to work. Can I open a PR to update the `run` target ? 

If yes do you need additional tests for it ?

PR #715, opened 2017-08-09
Convert the run target to double-colon

benjamin-bergia:
Related to #714 

Convert the run target to double-colon. This way the run target can be extended by users.

### DIFF
--- a/plugins/relx.mk
+++ b/plugins/relx.mk
@@ -54,7 +54,7 @@ distclean-relx-rel:
 # Run target.
 
 ifeq ($(wildcard $(RELX_CONFIG)),)
-run:
+run::
 else
 
 define get_relx_release.erl
@@ -78,7 +78,7 @@ ifeq ($(PLATFORM),msys2)
 RELX_REL_EXT := .cmd
 endif
 
-run: all
+run:: all
 	$(verbose) $(RELX_OUTPUT_DIR)/$(RELX_REL_NAME)/bin/$(RELX_REL_NAME)$(RELX_REL_EXT) console
 
 help::


essen:
No tests needed, I have no idea how to test the run target at this point other than manually.

essen:
Merged, thanks!

nickelization:
My make skills are not super sharp, so I could be confused/mistaken, but I think this patch is causing some compatibility problems between different versions of erlang.mk.

I have a project that uses a recent version of erlang.mk with this change, but pulls in dependencies using older erlang.mk versions, and `make run` yields this error:
```
makefile:22: *** target file 'run' has both : and :: entries.  Stop.
```
Any good solution for this that doesn't involve upgrading every dependency to a newer erlang.mk version? In the mean time I downgraded my own project to an older erlang.mk and it does eliminate the problem, but of course that's just a temporary workaround. I am a bit worried that this issue could become difficult to manage if too many third-party dependencies end up using different versions that aren't compatible with each other.

essen:
Are you sure you don't just have a `run:` target in your Makefile that you can remove? If it only happens on `make run` then I doubt it concerns the dependencies.

essen:
You can also use the `NO_AUTOPATCH_ERLANGMK=1` variable to make the dependencies keep their own Erlang.mk. Results may vary though.

nickelization:
Aha...so I do not have a `run:` target in the top-level project with the newer erlang.mk, but one of the dependency projects does. So this change does break compatibility between my newer project and that version of the dep, but fortunately in this case the dependency is an internal project that I have full control over, so it's easy for me to upgrade and fix this.

So anyway, yeah, looks like this won't cause quite as many problems as I'd feared, and should only be an issue for projects that depend on other projects which define a `run:`/`run::` target in their Makefile. Not sure how common this is in practice, so no idea if it's a case that's really worth worrying about or not.

essen:
It's very rare for projects used as dependencies, I think i've only seen it once in the wild. More common in top-level projects that create the release.

PR #719, opened 2017-08-24
Use kerl environment variables

benjamin-bergia:
With this change, users can fetch releases directly by default and set `KERL_BUILD_BACKEND` to `git` in order to use the OTP github repo.


essen:
I would much prefer the git backend by default, it has better availability than erlang.org.

benjamin-bergia:
Didn't know about that. In this case this PR might not be very useful. 
I will test something different over the weekend.
If I can't get it to work I will close the PR.

benjamin-bergia:
I will close this. Passing environment variables can easily be done with the current version.

PR #721, opened 2017-08-30
Support in pattern matching of extra data type

plakshman93:
This will support defining macro in `rebar.config.script`. An extra data type pattern matching is added in `dep_autopatch_rebar` function.


essen:
Thanks, sounds good. Do you have an example project I could test against perhaps?

plakshman93:
yes. You can look at this  github repo.
https://github.com/tsloughter/erlastic_search.

essen:
This fixes the compilation of this project, nice! Merged, thanks!

PR #722, opened 2017-09-07
COVER_DATA_DIR for *.coverdata

nevar:
Also make COVER_REPORT_DIR not override user value (if set before include erlang.mk).
Use incl_app in CT cover spec.

### DIFF
--- a/plugins/cover.mk
+++ b/plugins/cover.mk
@@ -2,23 +2,20 @@
 # Copyright (c) 2015, Viktor Söderqvist <viktor@zuiderkwast.se>
 # This file is part of erlang.mk and subject to the terms of the ISC License.
 
-COVER_REPORT_DIR = cover
+COVER_REPORT_DIR ?= cover
+COVER_DATA_DIR ?= $(CURDIR)
 
 # Hook in coverage to ct
 
 ifdef COVER
 ifdef CT_RUN
 ifneq ($(wildcard $(TEST_DIR)),)
-# All modules in 'ebin'
-COVER_MODS = $(notdir $(basename $(call core_ls,ebin/*.beam)))
-
 test-build:: $(TEST_DIR)/ct.cover.spec
 
-$(TEST_DIR)/ct.cover.spec:
-	$(verbose) echo Cover mods: $(COVER_MODS)
+$(TEST_DIR)/ct.cover.spec: cover-data-dir
 	$(gen_verbose) printf "%s\n" \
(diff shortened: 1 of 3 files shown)


essen:
A quick look tells me it's good but can you remind me on Monday? I won't have much time until then.

nevar:
Not sure that that for you already Monday, but ping!

plugins/eunit.mk
```diff
@@ -45,10 +45,10 @@ EUNIT_ERL_OPTS += -pa $(TEST_DIR) $(DEPS_DIR)/*/ebin $(APPS_DIR)/*/ebin $(CURDIR
 
 ifdef t
 ifeq (,$(findstring :,$(t)))
-eunit: test-build
+eunit: test-build $(if $(COVER),,| $(COVER_DATA_DIR))
```
essen: Rather than do this everywhere, I would rather you depend on a rule in cover.mk that either does nothing or depends on ` | $(...)` depending on COVER.

essen:
plugins/ct.mk didn't need any changes?

nevar:
Remove "| $(...)" and add rule for it in plugin/cover.mk.
All needed change for ct in plugins/cover.mk (in ct.cover.spec file template).
Also noticed that $(COVER_MODS) in plugin/cover.mk not used. Remove it?

essen:
If it's no longer necessary, yes, definitely remove.

nevar:
Remove $(COVER_MODS) and change message when creating ct.cover.spec

plugins/cover.mk
```diff
@@ -2,23 +2,21 @@
...
 ifneq ($(wildcard $(TEST_DIR)),)
-# All modules in 'ebin'
-COVER_MODS = $(notdir $(basename $(call core_ls,ebin/*.beam)))
-
 test-build:: $(TEST_DIR)/ct.cover.spec
 
-$(TEST_DIR)/ct.cover.spec:
-	$(verbose) echo Cover mods: $(COVER_MODS)
+$(TEST_DIR)/ct.cover.spec: | $(COVER_DATA_DIR)
+	$(verbose) echo CT cover spec for application $(PROJECT)
```
essen: Does the echo provides any useful information to the user? If not, just remove it.

Also use the cover-data-dir rule for this target also and then we should be good.

nevar:
Remove echo and replace all "| $(...)" to cover-data-dir.

essen:
I think we're good, I'll try to merge tomorrow.

essen:
Merged, thanks!

PR #723, opened 2017-09-11
Add rebar3 hex package define

nevar:
Able to compile elvis_core (master).

### DIFF
--- a/core/deps.mk
+++ b/core/deps.mk
@@ -271,6 +271,7 @@ define dep_autopatch_rebar.erl
 			{_, Deps} ->
 				[begin case case Dep of
 							{N, S} when is_atom(N), is_list(S) -> {N, {hex, S}};
+							{_, S, {pkg, N}} -> {N, {hex, S}};
 							{N, S} when is_tuple(S) -> {N, S};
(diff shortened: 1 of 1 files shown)


essen:
I have some doubts about this.

First, elvis_core doesn't have this in the normal dependencies, it's a test dependency and Erlang.mk ignores those: https://github.com/inaka/elvis_core/blob/master/rebar.config#L25

Second, I do not think that using the package name is the right thing to use because the application name is `mixer` and yet this would cause Erlang.mk to put it in `deps/inaka_mixer` and this will create some problems down the line (folder names are expected to be the same as the application's name by some things in Erlang).

This is related to https://github.com/ninenines/erlang.mk/pull/500

nevar:
But katana have thous type dep in deps [here](https://github.com/inaka/katana-code/blob/master/rebar.config#L70). So whitout this PR katana_code fail to build => elvis_code fail to build => ...

essen:
Ah I see. There's still the issue of the application name, so I think #500 needs to be solved before this.

essen:
Merged locally, sorry it took so long! Will push soon.

PR #727, opened 2017-10-06
Add package escalus to the index

oliv3:


essen:
I have merged it even though it doesn't compile, that's due to exml (at least on my machine). I suppose you have some workaround. Cheers.

PR #729, opened 2017-10-26
Fix missing core-misc target in tests

tomjnixon:
This updates core_misc.mk to match the other core_*.mk files, whose tests are included via the include_core mechanism in Makefile.

It's not clear to me why this would have worked for anyone or else (or the CI server), so appologies if i've missed something.

Thanks,

Tom

### DIFF
--- a/test/core_misc.mk
+++ b/test/core_misc.mk
@@ -5,9 +5,9 @@
 CORE_MISC_CASES = clean-crash-dump distclean-tmp help without-edoc without-index without-many
 CORE_MISC_TARGETS = $(addprefix core-,$(CORE_MISC_CASES))
 
-.PHONY: $(CORE_MISC_TARGETS)
+.PHONY: core-misc $(CORE_MISC_TARGETS)
 
-core:: $(CORE_MISC_TARGETS)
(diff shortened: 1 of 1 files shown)


essen:
You're not fixing anything here. :-)

The `::` means the target can be repeated to do extra things.

tomjnixon:
Well, it causes a failure on my machine because [this](https://github.com/ninenines/erlang.mk/blob/master/test/Makefile#L106-L113) causes `core` to depend on `core-misc`, which did not exist (as far as i can see, anyway)

tomjnixon:
Exact error message, from a clean checkout of master:

    $ make check
    ...
    core-makedep-import: Confirm the file was added by makedep
    make[1]: *** No rule to make target 'core-misc', needed by 'core'.  Stop.
    make[1]: Leaving directory '/home/tom/dev/erlang.mk/test'
    make: *** [Makefile:50: check] Error 2

I'm running archlinux.

essen:
Me too. I will check again and add it if necessary. It's possible something changed in Make recently.

essen:
Merged locally, thanks!

PR #730, opened 2017-10-26
Fix switched UID/GID in asciidoc install

tomjnixon:
This caused the `asciidoc-install` test to fail with this message on my system, where UID != GID:

    install: cannot change ownership of 'installed/share/man3/erlang_mk.3.gz': Operation not permitted

I'm not sure why the user and group flags were there in the first place, as this looks like the default for install.

Thanks,

Tom


essen:
Cheers I will take a look soon.

essen:
Merged locally, thanks!

PR #731, opened 2017-10-26
Use dep names and ignore IGNORE_DEPS in .app files

tomjnixon:
If an app depends on one of the packages whose package name does not match the application name (exec/erlexec and opencouch/couch), then the generated .app file will have the wrong name in the `applications` field.

It seems like `IGNORE_DEPS` should be followed here too, though that's less clear. This is basically the same as in the `ALL_DEPS_DIRS` variable.

Thanks,

Tom

### DIFF
--- a/core/erlc.mk
+++ b/core/erlc.mk
@@ -69,7 +69,7 @@ define app_file
 	{id$(comma)$(space)"$(1)"}$(comma))
 	{modules, [$(call comma_list,$(2))]},
 	{registered, []},
-	{applications, [$(call comma_list,kernel stdlib $(OTP_DEPS) $(LOCAL_DEPS) $(DEPS))]},
+	{applications, [$(call comma_list,kernel stdlib $(OTP_DEPS) $(LOCAL_DEPS) $(foreach dep,$(DEPS),$(call dep_name,$(dep))))]},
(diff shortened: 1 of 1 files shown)


essen:
No IGNORE_DEPS is not necessary here because it is meant for dependencies of dependencies (it makes no sense to define IGNORE_DEPS when you are the one who defines DEPS). Perhaps the behavior could be better documented though, I don't know.

tomjnixon:
Ah, that makes sense. I've updated the original commit, and but it might be neater to just inline it now, as in https://github.com/tomjnixon/erlang.mk/commit/52b49d1746d15fceac51192519940ba3dabaf799

essen:
Yeah let's go for the inline version. Thanks!

tomjnixon:
Great, thank you!

essen:
Merged locally. I still have errors from opencouch but that's because a curl call fails for some reasons. But since exec works fine now I'm fine with it. Thanks!

tomjnixon:
Thanks for all the merges! I didn't see the opencouch error when I tested this, so that sounds like it's probably new.

essen:
Yeah it doesn't sound related.

PR #732, opened 2017-10-30
Add output directory for generated EDoc

dozzie:


essen:
OK merged locally. Test would be better to have but current tests pass anyway. Thanks!

PR #735, opened 2017-11-25
Add grpc and grpc_client

willemdj:
gRPC server using cowboy HTTP2 implementation, and client to go with it.


essen:
Merged, thanks! I'll push in a few.

PR #742, opened 2018-02-07
Added support for ASN.1 compiler option "maps".

vanrein:
(The option generates no include/*.hrl files, which caused an error.)

I did not alter the erlang.mk, as that seems to be intended for local overwrite only.

I have tested this on Mac OS X, which is pretty much stock Unix on this level.


vanrein:
not good enough yet, sorry.  will use GNU makefile loops and update.

PR #744, opened 2018-02-11
Create temporary directory when generating relx

KrzysiekJ:


essen:
I disagree on the test changes, this particular issue should be a separate test. The tests mirror what most users would do so it's important to have that tested.

KrzysiekJ:
Ok, I’ve created a separate test.

essen:
Merged, thanks!

PR #751, opened 2018-03-13
Speed up the download done as part of bootstrapping.

### LINKED ISSUE #750: Use --depth in erlang.mk, git clone command to speed up download.
The current `erlang.mk` as obtained from [https://erlang.mk/erlang.mk](https://erlang.mk/erlang.mk)  clones the entire erlang.mk repo, including all its history, uses it for one call, and then deletes the whole repo.

Using the switch `--depth 1` considerably speeds this up, as not the entire history is downloaded.

aknrdureegaesr:
Fixes #750 .

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -78,7 +78,7 @@ docs:
 	$(MAKE) -f core/core.mk -f core/docs.mk -f plugins/asciidoc.mk asciidoc DEPS=asciideck
 
 up:
-	git clone git@github.com:ninenines/erlang.mk.git gh-pages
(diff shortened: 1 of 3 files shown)


essen:
Also needed for erlang-mk target in core/core.mk.

essen:
Merged and update erlang.mk/erlang.mk.

The change in the Makefile was incorrect so I fixed it again. It's something only used internally anyway. My bad.

essen:
Thanks!

Issue #757, opened 2018-03-28
.erl file is not recompiled if it's uses header from another application and that header is updated


stchar:
Multiapps project. One application uses headers of another. For example `my_app` uses headers from `my_lib` by `-include_lib() option`. LOCAL_DEPS of `my_app` includes `my_lib`
```
├── apps
│   ├── my_app
│   │   ├── Makefile
│   │   └── src
│   │       ├── use_blue.erl
│   │       └── use_red.erl
│   └── my_lib
│       ├── include
│       │   ├── blue.hrl
│       │   └── red.hrl
│       ├── Makefile
│       └── src
├── erlang.mk
└── Makefile
```
If user changes a header file in `my_lib`  (e.g. `my_lib/include/red.hlr`) it is expected that `use_red.erl` would be recompiled.

essen:
Good find. Tough one.

stchar:
Do you have any idea where to start the fix?

essen:
Not sure. Probably needs to add the dependency in the $PROJECT.d file but then this might not be the only issue.

stchar:
Yeah, if I manually add explicitly  rules to `apps/my_app/my_app.d` it does the job
```
# file apps/my_app/my_app.d
src/use_blue.erl:: ../my_lib/include/blue.hrl; @touch $@
src/use_red.erl:: ../my_lib/include/red.hrl; @touch $@

COMPILE_FIRST +=
```
Make output 
```
make[1]: Entering directory '/home/ovcharsa/git/hcm/erlang.mk/test/test_core_app_hrl_multiapps_include_lib/apps/my_app'
make[2]: Entering directory '/home/ovcharsa/git/hcm/erlang.mk/test/test_core_app_hrl_multiapps_include_lib/apps/my_lib'
 APP    my_lib
make[2]: Leaving directory '/home/ovcharsa/git/hcm/erlang.mk/test/test_core_app_hrl_multiapps_include_lib/apps/my_lib'
 DEPEND my_app.d
 ERLC   use_red.erl
 APP    my_app
```

essen:
I guess this is fixed once I push. Thanks!

PR #758, opened 2018-03-28
Recompile erl if hrl from another application it uses was updated

stchar:
Add a test scenarios for https://github.com/ninenines/erlang.mk/issues/757

### DIFF
--- a/test/core_app.mk
+++ b/test/core_app.mk
@@ -1,6 +1,6 @@
 # Core: Building applications.
 
-CORE_APP_CASES = appsrc-change asn1 auto-git-id env erlc-exclude erlc-opts erlc-opts-filter error extra-keys generate-erl generate-erl-include generate-erl-prepend hrl hrl-recursive makefile-change mib name-special-char no-app no-makedep project-mod pt pt-erlc-opts xrl xrl-help xrl-include yrl yrl-header yrl-include
+CORE_APP_CASES = appsrc-change asn1 auto-git-id env erlc-exclude erlc-opts erlc-opts-filter error extra-keys generate-erl generate-erl-include generate-erl-prepend hrl hrl-recursive makefile-change mib name-special-char no-app no-makedep project-mod pt pt-erlc-opts xrl xrl-help xrl-include yrl yrl-header yrl-include hrl-include-lib hrl-include-lib-recursive
 CORE_APP_TARGETS = $(addprefix core-app-,$(CORE_APP_CASES))
 
 .PHONY: core-app $(CORE_APP_TARGETS)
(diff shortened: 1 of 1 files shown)


essen:
Cheers. I'll see when I can make time for it.

core/erlc.mk
```diff
@@ -185,8 +186,14 @@ define makedep.erl
...
 			end;
-		(F, Mod, include_lib, "$1/include/" ++ Hrl) -> AddHd(F, Mod, "include/" ++ Hrl);
-		(F, Mod, include_lib, Hrl) -> AddHd(F, Mod, "include/" ++ Hrl);
+		(F, Mod, include_lib, "$1/include/" ++ Hrl) ->
+			AddHd(F, Mod, "include/" ++ Hrl);
+		(F, Mod, include_lib, Hrl) ->
+			HrlFile = filename:join(["$(APPS_DIR)",Hrl]),
+			case filelib:is_file(HrlFile) of
+				true -> AddHd(F, Mod, HrlFile);
+				false -> AddHd(F, Mod, "include/" ++ Hrl)
```
essen: If the file doesn't exist I'm not sure we should do this at all.

stchar: It's to support "local" headers in `apps/app_dir/include/some_header.hlr`
So this chunk would work
```erlang
-include_lib("some_header.hlr")
```

essen:
No objections but one concern. This takes care of apps/, what about deps/? I'm guessing we'll run into the same problems there.

core/erlc.mk
stchar: hmm. I think the behavior should be the same like with `-include(File)`
* check `apps/+Hlr`
* check `include/+Hlr`
* check `src/+Hlr`

essen: Perhaps the same algorithm used in the Erlang compiler can be reused for this?

stchar:
Do you mean "BUILD_DEPS?", some applications we load from outside. 
If so their sources should not be changed and should be compile just once if needed. Well IMHO.

essen:
DEPS or BUILD_DEPS, same thing in this context. They can change if you're working on both the app and a dep (say, cowboy+cowlib).

stchar:
Updated to support `DEPS_DIR`
I think It also needs adding test scenarios on header location preemption
* `src/` (none)
* `include/` (core-app-hrl-include-lib-*)
* `apps/` (core-app-hrl-multiapps-include-lib-*)
* `deps/` (none) I saw something similar in `test/core_deps.mk` with cowlib

core/erlc.mk
```diff
@@ -185,8 +186,29 @@ define makedep.erl
 						false -> false
 					end
 			end;
-		(F, Mod, include_lib, "$1/include/" ++ Hrl) -> AddHd(F, Mod, "include/" ++ Hrl);
-		(F, Mod, include_lib, Hrl) -> AddHd(F, Mod, "include/" ++ Hrl);
+		(F, Mod, include_lib, "$1/include/" ++ Hrl) ->
```
stchar: i'm not sure if we need line 189

essen:
Thanks. And yeah I suppose it's missing a lot of tests. If you have time please try to add them.

test/core_app.mk
```diff
@@ -1758,3 +1758,391 @@ endif
...
+	$t printf "%s\n" \
+		$(APP)/apps/my_app/my_app.d \
+		$(APP)/apps/my_app/ebin/my_app.app \
+		$(APP)/apps/my_app/ebin/use_red.beam \
+		$(APP)/apps/my_app/src/use_red.erl | sort > $(APP)/EXPECT
+	$t $(SLEEP)
+	$t touch $(APP)/apps/my_lib/include/pill.hrl
+	$t $(SLEEP)
+	$t $(MAKE) -C $(APP) $v
+	$t find $(APP)/apps/my_app -type f -newer $(APP)/apps/my_lib/include/pill.hrl | sort #| diff $(APP)/EXPECT -
```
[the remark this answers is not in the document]
stchar: If I'm not mistaken `make` for gcc have the same issue. 
https://stackoverflow.com/questions/32784369/how-to-use-gcc-m-for-autodependency-generator

I believe this should be changed to use `-M` & `-MF` of `erlc` as it looks close to gcc-way
http://erlang.org/doc/man/erlc.html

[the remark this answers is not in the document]
essen: I think `erlc -M` was not giving me what I wanted but you're welcome to experiment.

core/erlc.mk
stchar: seems we do not, 
looks it's not covered by any testcase.

test/core_app.mk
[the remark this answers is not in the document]
stchar: Ok, let me fix the bug first.
I heave some ideas, I would try it on a separate branch.

stchar:
Lol, I just realize that `makedep.erl` has no protection from include loop.
when foo.hlr includes bar.hlr while bar includes foo.

People use macro guards like to escape this loop, however `io:parse_erl_form` fails to process `ifndef ` macro
```erlang
-ifndef(FOO_ALREADY_INCLUDED).
-define(FOO_ALREADY_INCLUDED,true)
%% header body %%
-endif.
```

essen:
Not terribly surprised hehe. Thanks for taking a good look into this.

core/erlc.mk
```diff
@@ -146,6 +146,8 @@ define makedep.erl
 	E = ets:new(makedep, [bag]),
 	G = digraph:new([acyclic]),
 	ErlFiles = lists:usort(string:tokens("$(ERL_FILES)", " ")),
+	DepsDirs = lists:usort(string:tokens("$(wildcard $(DEPS_DIR)/*/src) $(wildcard $(DEPS_DIR)/*/include)", " ")),
```
stchar: On windows `$(wildcard $(DEPS_DIR)/*/src)` produces path like `/home/user/erlang.mk/...`
`filelib:is_file("/home/user/...")` returns false even if the file exists as erlc cannot determine if it is in msys2 environment. 

To fix it on windows msys2 paths should transformed whether into mixed type of whether into windows: 
* `/home/user/` should be replaced as `C:\\tools\\msys2\\home\\user`
* `/home/user/` should be replaced as `c:/tools/msys2/home/user`

essen: Right. `core_native_path` it is.

essen:
What's the progress on this, is it ready for review?

stchar:
Yes, it is. I've tested it on Linux and win10, haven't got any Mac in my build environment.

essen:
I've merged locally and I will push in a few, thanks!

stchar:
Cool, thanks.

Issue #759, opened 2018-03-29
Q: Vagrantbox/Dockerfile to run erlang.mk tests


stchar:
@essen,
Erlang.mk includes plugins which may require users to install additional system packages.
So I'm wondering if you have any draft solutions to describe test environment.

essen:
`erlang git gmake autoconf automake gcc` should get you most of the way. It's possible something is missing I am unfortunately not doing a good job of maintaining that list. Feel free to point me to failing tests and I'll guide you toward the packages necessary.

stchar:
I have some tests that does not work. 
The some look outdated, some broken,  I can delete/comment-out them in a separate commit.
BTW how do you run any CI builds for erlang.mk PRs?

essen:
I have permanent Arch Linux, Ubuntu, FreeBSD, OSX (and previously, Windows, temporarily halted due to bs) on real servers and/or VMs. Some BuildKite configuration for what to run (below) and its agent does the rest.

```
make check -j 8 -k
make check LEGACY=1 -j 8 -k
```

essen:
Oh and note that when you push it doesn't currently run the tests for PRs, it's behind a lock and I have to click it manually to make it really run. So if it says successful here it doesn't mean the tests ran successfully.

stchar:
Yep, I noticed that. 
That's why I'd like to have a kind of dev env to verify that my PR does't break the regression.

essen:
Yep no worries I'll test the PR soon.

stchar:
`-k ` means keep going, so you run as much tests as possible.
How do you analyze what test failed?

stchar:
```
grep -i error
```

essen:
Actually yes that's how. Though I do this on the output log file rather than via pipes (Erlang + pipes is not the best combination).

stchar:
cool, thanks/

PR #760, opened 2018-04-02
Add dockerfile to support development on linux

stchar:
https://github.com/ninenines/erlang.mk/issues/759

### DIFF
--- a/docker/Dockerfile
+++ b/docker/Dockerfile
@@ -1,0 +1,24 @@
+FROM ubuntu
+
+RUN apt-get update && apt-get install -y \
+  git \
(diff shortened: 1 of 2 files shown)


essen:
Man it takes forever to download all these packages. Seems good, thanks!

stchar:
The very first time it is a pain or when you update dependencies list. But then docker takes images from local cache. 
Usually they use 2 docker files. 
* 1st is to describe the environment. 
* 2nd is based on the 1st and is used for daily activities (compilation, test running). 

1st image could pushed to the docker registry so people can download it ready to go.

PR #764, opened 2018-04-06
Fix rebar pre_hooks patch

### LINKED ISSUE #763: erlang.mk incorrectly patch rebar pre_hooks
When I try to use this [dependency](https://github.com/Ledest/otpbp), I get error message:

> make[2]: Entering directory '/home/kostyushkin/sandboxes/ecache/deps/otpbp'
> CC=cc make -f Makefile.orig.mk -s -C src/uc_spec unicode_util.erl
> make[3]: Entering directory '/home/s.kostyushkin/sandboxes/ecache/deps/otpbp/src/uc_spec'
> make[3]: Makefile.orig.mk: No such file or directory
> make[3]: *** No rule to make target 'Makefile.orig.mk'.  Stop.

It looks like the issue is here: [deps.mk#L337](https://github.com/ninenines/erlang.mk/blob/master/core/deps.mk#L337)
The problem occurs because real life rebar [pre_hooks](https://github.com/Ledest/otpbp/blob/master/rebar.config.script#L199) does not match this pattern.

kostyushkin:
Fix #763

### DIFF
--- a/core/deps.mk
+++ b/core/deps.mk
@@ -332,7 +332,8 @@ define dep_autopatch_rebar.erl
 	Write("\npre-deps::\n"),
 	Write("\npre-app::\n"),
 	PatchHook = fun(Cmd) ->
-		case Cmd of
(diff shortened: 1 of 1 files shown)


essen:

PR #766, opened 2018-04-13
document PROJECT_ENV (and file) usage

jimdigriz:
Without spelunking through erlank.mk, it is not clear how to use `PROJECT_ENV`, especially if you wish to use a file instead.


essen:

PR #771, opened 2018-04-23
add gen_statem template

juhlig:
a template for gen_statem
updated bootstrap test to include it
updated getting started documentation to include it


essen:
Merged, thanks!

PR #784, opened 2018-05-23
Add relx-post-rel hook

lego12239:
This hook can be used, for example, to copy additional
files to release before it is tared by relx.


essen:
Yes, indeed, after release and before relup.

PR #785, opened 2018-05-23
Add relx-post-rel hook

lego12239:
This hook can be used, for example, to copy additional
files to release before it is tared by relx.

### DIFF
--- a/plugins/relx.mk
+++ b/plugins/relx.mk
@@ -44,13 +44,22 @@ $(RELX):
 	$(verbose) chmod +x $(RELX)
 
 relx-rel: $(RELX) rel-deps app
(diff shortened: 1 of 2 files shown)


lego12239:
Please check relx-relup target.

essen:
Yep looks fine. Can you add a test?

essen:
By the way you don't need to close/reopen the PR every time, you can just `git push -f` your branch and it'll update.

lego12239:
> Can you add a test?

Hm. I can't do this quickly :-). I need to understand the test suite.

essen:
Merged locally, push later today. Thanks!

PR #787, opened 2018-05-29
Add $(TEST_DIR) to the default value of SHELL_PATHS

KrzysiekJ:


essen:
Should probably have a test case, you can copy paste `shell-default` and ensure it works by calling a function from the test directory at the end.

KrzysiekJ:
Ok, uploading a version with test.

essen:
Merged, thanks!

PR #792, opened 2018-06-05
Use 'Noninteractive' frontend in Dockerfile

rlipscombe:
Otherwise it stops to ask what timezone you want to use.


essen:

PR #799, opened 2018-07-01
Swallow warnings when creating a PLT

### LINKED ISSUE #501: make dialyze problem with erlydtl as dependency
When adding erlydtl as a dependency, `make plt` complains:

```
erlydtl_compiler.erl:429: Call to missing or unexported function erlydtl_parser:resume/1
erlydtl_filters.erl:543: Call to missing or unexported function erlydtl_time_compat:phash2/1
```

And exits with code 2, failing the build. This is due to code generated from yecc.

However, the plt is created anyway and a second invocation to `make dialyze` (which depends on the plt) still succeed if the main app is free of error.

The problem is: this makes running `make check` or `make dialyze` in a CI environment quite inconvenient. One would have to `make plt` and swallow the error first.

I'm not sure what is the correct way to fix this. Specific autopatch for erlydtl? Correctly handle generated code? Ignore error from `plt` target if the plt is built anyway (i.e: `rm $(DIALYZER_PLT) && dialyzer --build_plt ... || [ -f $(DIALYZER_PLT) ]`?

KrzysiekJ:
Fixes #501.


essen:
Merged locally, thanks!

PR #801, opened 2018-07-13
Remove kakfa_protocol from index

zmstone:


essen:
Any reasons why?

zmstone:
The latest master branch no longer supports erlang.mk (rebar3 only).

essen:
It still builds so I'm not seeing a problem with that.

zmstone:
ok. declined. thanks!

Issue #802, opened 2018-07-14
Generated parser file not found when switching between test mode and non-test mode


darkling:
```
$ make
[... success ...]
$ make check            # The first time fails
 GEN    clean-app
 GEN    coverdata-clean
make[1]: Entering directory '/home/hrm/nightglass/software/childe/deps/erlsom'
[...]
 ERLC   childe.erl childe_app.erl childe_cell_formatter.erl childe_data_load.erl childe_instance.erl childe_lexer.erl childe_logger.erl childe_parser_tests.erl childe_pi.erl childe_pretty.erl childe_pretty_gen.erl childe_proc.erl childe_report_eval.erl childe_report_formatter.erl childe_roster.erl childe_semantic.erl childe_sup.erl childe_trace.erl childe_types.erl childe_parser.erl
src/childe_parser.erl: no such file or directory
make[1]: *** [erlang.mk:5228: ebin/childe.app] Error 1
make: *** [erlang.mk:5337: test-build] Error 2
$ make check            # The second time succeeds
 GEN    clean-app
 GEN    coverdata-clean
make[1]: Entering directory '/home/hrm/nightglass/software/childe/deps/erlsom'
[...]
 XYRL   childe_parser.yrl
[...]
 DEPEND childe.d
 ERLC   childe.erl childe_app.erl childe_cell_formatter.erl childe_data_load.erl childe_instance.erl childe_lexer.erl childe_logger.erl childe_parser.erl childe_parser_tests.erl childe_pi.erl childe_pretty.erl childe_pretty_gen.erl childe_proc.erl childe_report_eval.erl childe_report_formatter.erl childe_roster.erl childe_semantic.erl childe_sup.erl childe_trace.erl childe_types.erl
 APP    childe
 GEN    test-dir
[... success ...]
```

And similarly going in the other direction -- from this state, the first run of ```make``` fails with ```src/childe_parser.erl: no such file or directory```. The second run succeeds.

Using the most recent erlang.mk. 

[Makefile.txt](https://github.com/ninenines/erlang.mk/files/2195009/Makefile.txt)

essen:
In that situation the actual clean should probably occur in a separate make invocation (erlang.mk can detect and act accordingly). Can you perform a small experiment where you put `app:: ; $(MAKE) clean` at the beginning of your Makefile when you switch from one to the other to see if it works like this?

darkling:
No, that doesn't seem to help.

```
$ make distclean
[...]
$ make ct
[...]

PR #803, opened 2018-07-14
Test case for yecc, switching between test mode and normal mode

darkling:
Re issue #802 

With a yecc grammar in the project, and this sequence of commands:

$ make clean
$ make check
$ make
$ make

the "make check" succeeds. The first "make" fails because the .erl
file built from the .yrl file doesn't exist. The second "make"
succeeds.

Add a test case for this situation.

Signed-off-by: Hugo Mills <hugo@nightglass.co.uk>

### DIFF
--- a/test/core_app.mk
+++ b/test/core_app.mk
@@ -1758,6 +1758,26 @@ endif
 		[{module, M} = code:load_file(M) || M <- Mods], \
 		halt()"
 
+core-app-yrl-normal-to-debug: build clean
+
+	$i "Bootstrap a new OTP library named $(APP)"
+	$t mkdir $(APP)/
+	$t cp ../erlang.mk $(APP)/
+	$t $(MAKE) -C $(APP) -f erlang.mk bootstrap-lib $v
+
+	$i "Create the parser .yrl file"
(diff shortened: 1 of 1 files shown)


test/core_app.mk
```diff
@@ -1758,6 +1758,26 @@ endif
...
 
+core-app-yrl-normal-to-debug: build clean
+
+	$i "Bootstrap a new OTP library named $(APP)"
+	$t mkdir $(APP)/
+	$t cp ../erlang.mk $(APP)/
+	$t $(MAKE) -C $(APP) -f erlang.mk bootstrap-lib $v
+
+	$i "Create the parser .yrl file"
+	$t echo "Nonterminals top.\nTerminals plus num.\nRootsymbol top.\ntop -> num plus num : {'$1', '$3'}." >$(APP)/src/$(APP)_parser.yrl
```
essen: You can't use echo for inserting new lines, its behavior depends on the shell and platform. Please use `printf "%s\n" "line1" "line2..."` instead.

essen:
I've got a fix for this but I'll probably push tomorrow, gotta wait for all tests to run.

PR #824, opened 2019-02-01
adding eunit setup capability

korroktheslavemaster:
Eg. I want to start an mnesia instance before all the tests and load my schemas. Then my tests can use this as the test db without having to painfully mock db calls.

For this I define in my own Makefile:
```
include erlang.mk
...
# setting up mnesia for unit tests
define eunit_setup.erl
	EunitSetup = fun () -> 
		application:start(mnesia),
                 %% loading my schemas/tables...
		end,
	EunitTeardown = fun () -> application:stop(mnesia) end,
endef
```
If there's a better solution please help me out, thanks!


essen:
EUnit already has setup/cleanup fixtures, see http://erlang.org/doc/apps/eunit/chapter.html#fixtures

Can't that be used instead?

korroktheslavemaster:
Yes, but that is at a module level. I would still have to setup mnesia and load schemas for each module. That is too expensive since I have hundreds of modules, and loading schemas takes quite a bit of time.

essen:
No I don't think it is per module.

In our `eunit.erl` the argument `$1` is the tests and can be a single module, a specific test function or all modules depending on what argument you give to Erlang.mk. If what I'm reading on that link is correct, then instead of giving, say, `[a, b, c]` to test those 3 modules, you could give `{setup, Setup, Cleanup, [a, b, c]}` and the `Setup` function would be executed before the tests of the 3 modules, then the `Cleanup` would be executed.

Maybe Erlang.mk needs a way to configure those fixture things. The `setup` tuple seems to run before/after the entire set of tests, while the `foreach` tuple seems to run before/after each test. So in your case the `setup` tuple should do what you want.

korroktheslavemaster:
Actually you might be right, let me look into it once more

essen:
If I go to my Cowlib project and edit the erlang.mk line where `eunit:test` is called and change it to:

```
case eunit:test({setup, fun() -> erlang:display("setup") end, fun(_) -> erlang:display("cleanup") end, $1}, [$(EUNIT_OPTS)]) of
```

I get the following output:

```
% make eunit
make[1]: Entering directory '/home/essen/ninenines/cowlib/deps/horse'
 DEPEND horse.d
 ERLC   horse.erl horse_autoexport.erl
 APP    horse.app.src
make[1]: Leaving directory '/home/essen/ninenines/cowlib/deps/horse'
make[1]: Entering directory '/home/essen/ninenines/cowlib/deps/proper'
 DEPEND proper.d
make[1]: Leaving directory '/home/essen/ninenines/cowlib/deps/proper'
 DEPEND cowlib.d
make[1]: Nothing to be done for 'test-dir'.
 GEN    eunit
"setup"
"cleanup"
  All 563 tests passed.
```

plugins/eunit.mk
```diff
@@ -20,10 +20,18 @@ help::
...
+	EunitGlobalTeardown = fun (_) -> ok end,
+endef
+endif
+
 define eunit.erl
 	$(call cover.erl)
+	$(call eunit_global_setup.erl)
 	CoverSetup(),
-	case eunit:test($1, [$(EUNIT_OPTS)]) of
+	case eunit:test({setup, EunitGlobalSetup, EunitGlobalTeardown, $1}, [$(EUNIT_OPTS)]) of
```
essen: Probably works better as:

```
case eunit:test($(call eunit_test_spec.erl,$1), [$(EUNIT_OPTS)]) of
```

And then you'd have by default:

```
EUNIT_TEST_SPEC ?= $1
```

Then in your Makefile you could define:

```
EUNIT_TEST_SPEC = {setup, fun my_tests:setup/0, fun my_tests:cleanup/1, $1}
```

Didn't test but that should be close to what we want. Then users can use this variable to use `foreach` or `foreachx` as well or make more complex test scenarios, and they can then have separate targets for separate `EUNIT_TEST_SPEC` values if necessary.

essen: I'll just have to double check the variable name is good before merging, I'm not sure we have too many cases of a configurable `$(call ...)` at the moment.

korroktheslavemaster:
made changes

essen:
Merged and added a test. Thanks!

PR #825, opened 2019-02-12
Update bootstrap.mk

marcinurbanski:
Add possibility to use project with dash between (foo-project1)


essen:
That's wrong because when Erlang sees a folder called `foo-project1` in `ERL_LIBS` it thinks the application is named `foo` and `project1` is its version.

marcinurbanski:
So is it not possible to use dash separated project name?

essen:
Not to my knowledge. Some things will work, some will break. Use underscores instead.

marcinurbanski:
Ok. Thanks for quick explanation. Have nice day!

essen:
Enjoy!

PR #827, opened 2019-03-01
Adding gen_leader and gen_rpc packages

stephb9959:
Just adding gen_leader and gen_rpc packages.


essen:
Merged, thanks! Note that `hut` dependency of `gen_rpc` fails to compile, you might want to move the `ci` target at the end of the Makefile. (Maybe the `app` target as well.)

PR #835, opened 2019-05-15
Fixed issue with {vsn,git}.  Added support for the fossil sources.

ad-pro:


essen:
I've done the change manually since the PR had two extra commits. Thanks!

PR #836, opened 2019-05-22
Fossil

ad-pro:
Update support of Fossil repository. Work on Windows.


essen:
Nice! Could you add a test in test/core_deps.mk? Perhaps copying the one for git.

PR #838, opened 2019-06-17
Set locale to `C` when checking bootstrapped app name

dumbbell:
At least on FreeBSD, the `en_US.UTF-8` locale I use breaks the check. It might be a bug on the FreeBSD 13-CURRENT I'm using, because I don't understand why it would affect the behaviour of that simple regex. Nonetheless, let's override the locale to make sure this does not happen again.


essen:
Merged, thanks!

PR #839, opened 2019-06-17
Replace zipinfo(1) by `unzip -l`

dumbbell:
The output is the same between those two commands. The benefit is that it avoids the need to install the `unzip` package on FreeBSD and all systems where the default unzip(1) is provided by libarchive.


essen:
Merged, thanks!

PR #840, opened 2019-06-17
Include applications `$(APPS_DIR)` in the `list-deps` targets

dumbbell:
In the case of a multi-applications repository, if one application depends on another one (using `$(LOCAL_DEPS)`), it will now be listed in the `list-deps` targets.


essen:
Need a test that has `LOCAL_DEPS = ssl` or similar to see if it still works in that case (it probably does). Otherwise good. Maybe just add `ssl` to the existing test case.

essen:
Also please add the following to the changelog:

```
2019/06/20: `list-deps` now includes local applications
            found in the `APPS_DIR` directory.
```

dumbbell:
I addressed both comments.

essen:
Will merge tomorrow!

essen:
Merged, thanks!

PR #841, opened 2019-06-18
Use echo(1) instead of printf(1) in `core-autopatch-extended-erlc-opts`

dumbbell:
... to create the `autopatch-couchbeam` recipe in the testcase.

Newlines characters were interpreted by echo(1) in the `core-autopatch-extended-erlc-opts` recipe which was too early: the generated recipe was malformed.

Using echo(1) and its implicit appended newline character fixes the issue.


essen:
Merged, thanks!

PR #843, opened 2019-06-18
Display Hex dependency version in the "DEP mydep (version)" line

dumbbell:
... exactly as it is displayed for Git-based dependencies.


essen:
Merged, thanks!

PR #845, opened 2019-06-21
Test `list-deps` behavior with various values of `FULL`

dumbbell:
Here, we want to make sure that `list-deps` always lists all dependencies, no matter if they are up-to-date or not. I.e. we want to always visit them and list them recursively.


essen:
Merged, thanks!

PR #847, opened 2019-06-22
Use forward slashes in the Windows paths

essen:
Let's see if this breaks anything.


essen:
Oh yeah I don't need to open a PR for this.

PR #848, opened 2019-06-24
Accept `$(force_rebuilding_dep)` to tune the behavior of `$(FULL)`

dumbbell:
By default, Erlang.mk only builds dependencies once, the first time the top-level project is compiled. A forced rebuild of all dependencies can be forced by setting the `$(FULL)` variable to a non-empty value.

For developers working on the top-level project and one or more dependencies at the same time, Erlang.mk now accepts a variable/function called `$(force_rebuilding_dep)` which is called to determine if the dependency passed as the first argument should be forcibly rebuilt.

It allows a developer to filter the dependencies he works on (so they are always rebuilt) from other third-party dependencies he never touches (i.e. they are only built once at the beginning).

The content of `$(force_rebuilding_dep)` is expected to be a shell one-liner where the `$(1)` Make variable is set to the path to the dependency.

Here is an example from the testsuite:

```
force_rebuilding_dep = test '$(1)' = '$(CURDIR)/deps/cowlib'
```

### DIFF
--- a/core/deps.mk
+++ b/core/deps.mk
@@ -130,7 +130,7 @@ ifneq ($(ALL_DEPS_DIRS),)
 			:; \
 		else \
 			echo $$dep >> $(ERLANG_MK_TMP)/deps.log; \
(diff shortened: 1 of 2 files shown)


essen:
Merged, thanks!

PR #851, opened 2019-06-25
Accept a list of deps to forcibly rebuild in `$(FORCE_REBUILD)`

dumbbell:
... in addition to the `$(force_rebuild_dep)` function.

It's easier for the common use case than specify a function.


essen:
Can you also change the test where I used force_rebuild_dep to fix Windows in HEAD or HEAD^?

dumbbell:
Done!

essen:
Merged! Thanks.

This will be pushed when the previous build finishes, I want to see if Erlang.mk becomes green for the first time ever. Good chance that could be it!

PR #855, opened 2019-07-30
Fix concurrent creation of `recursive-*deps-list.log` files

dumbbell:
To do that, we simply include the PID of the Make process in temporary file.

We also compare the finished temporary file with the actual target file. If they are different, the temporary file is moved to the final target, otherwise, we just get rid of the temporary file. This way, the target timestamp stays the same if the content does not change.

### DIFF
--- a/core/deps-tools.mk
+++ b/core/deps-tools.mk
@@ -53,7 +53,7 @@ ifneq ($(filter shell,$(DEP_TYPES)),)
 $(ERLANG_MK_RECURSIVE_DEPS_LIST): $(ALL_SHELL_DEPS_DIRS)
 endif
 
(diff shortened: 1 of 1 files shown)


essen:
That's a nice trick you got there.

How did you get the failure to begin with? I don't think I've seen it in tests.

dumbbell:
In the RabbitMQ testsuite, when we try to start several nodes in parallel: the Make recipe starting a node depends on our `dist` or `test-dist` targets which in turn depends on those recursive deps list.

The actual error was:
* one instance removed the temporary file
* another one tried to sort its content -> the shell failed to open the just removed file

essen:
Merged, thanks!

Future PR runs should succeed on Windows, I've updated the configuration to be like the main repo runs.

PR #856, opened 2019-07-30
Recursive deps list: don't handle apps specifically

dumbbell:
They are already listed in `$^` and thus already taken care of. No need to visit them twice.


essen:
Merged, thanks!

Issue #858, opened 2019-08-12
Build symlinks


Bathtor:
I'm wondering if there's a way to make erlang.mk build symlinked files?

Use case:
I've got a folder with protocol buffer files I'm using across a bunch of projects in different languages, so I need to build them for each language separately. Clearly, I don't want to copy them everywhere, in case I make changes. But erlang.mk doesn't seem to build them if I symlink the global folder or the individual files into the `src/` folder.

Alternatively, a makefile option for additional source directories would also work for me, as long as it doesn't put the build products into my global proto folder.

Any help would be appreciated :)

Note: I'm using https://github.com/Bluehouse-Technology/grpc, which uses https://github.com/tomas-abrahamsson/gpb

essen:
Just to be clear the .proto files themselves are symbolic links?

In that case the culprit is probably https://github.com/ninenines/erlang.mk/blob/master/core/core.mk#L188 as it only looks for regular files.

If you can experiment and show me how to use `find` to get both regular and symbolic files in one invocation then I think we can add a test and fix it.

Bathtor:
I tried both linking directory and files, actually.
Either one would work for me if implemented.

On MacOS at least `find . \( -type l -o -type f \) -ls` seems to do the trick.

essen:
Ah yeah I forgot about `-o`. Can you add a test perhaps based on https://github.com/ninenines/erlang.mk/blob/master/test/core_app.mk#L327 except you don't need to set `ERLC_OPTS` and you can generate the files outside the `src/` directory and symlink them to mirror your use case. I don't think we need a .proto specific test, just .erl files should be fine.

Bathtor:
Sure, I'll look into that tomorrow and then do a PR.

PR #859, opened 2019-08-14
Build symlinked source files

Bathtor:
Addresses #858.

Contains both a test case and the change that makes the test pass.


essen:
Merged, thanks!

PR #860, opened 2019-08-14
Building .proto files with imports fails

Bathtor:
It seems that building protobuf files that have imports on other files in the same folder fails to find the imported file. Something is wrong in the path where it looks for the imports.

I added tests for both gpb and protobuffs, and both show the same behaviour.

In fact I managed to get it to work for gpb by putting the imported file also on the projects top level, because gpb looks for `./<importfile>.proto` and it seems the `./` is top level not `src/` when it is invoked.

It's probably an easy fix, I'm just terrible with Makefiles.

### DIFF
--- a/test/plugin_protobuffs.mk
+++ b/test/plugin_protobuffs.mk
@@ -40,6 +40,38 @@ protobuffs-compile: init
 		{ok, [empty_pb, simple_pb]} = application:get_key($(APP), modules), \
 		halt()"
 
+protobuffs-compile-imports: init
+
+	$i "Bootstrap a new OTP library named $(APP)"
+	$t mkdir $(APP)/
+	$t cp ../erlang.mk $(APP)/
+	$t $(MAKE) -C $(APP) -f erlang.mk bootstrap-lib $v
+
+	$i "Add protobuffs to the list of dependencies"
+	$t perl -ni.bak -e 'print;if ($$.==1) {print "BUILD_DEPS = protobuffs\n"}' $(APP)/Makefile
+
+	$i "Download two proto files with an import"
+	$t mkdir $(APP)/src/proto/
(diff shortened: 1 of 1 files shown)


essen:
Based on those two paragraphs we are doing it wrong: https://github.com/tomas-abrahamsson/gpb/blob/master/src/gpb_compile.erl#L268

Bathtor:
> Based on those two paragraphs we are doing it wrong: https://github.com/tomas-abrahamsson/gpb/blob/master/src/gpb_compile.erl#L268

Hmm, yeah, so the following change at least gets rid of the not found error for gpb:
```diff
diff --git a/plugins/protobuffs.mk b/plugins/protobuffs.mk
index b4f9ce6..d56c2e6 100644
--- a/plugins/protobuffs.mk
+++ b/plugins/protobuffs.mk
@@ -43,7 +43,8 @@ define compile_proto.erl
                        {include_as_lib, true},
                        {module_name_suffix, "_pb"},
                        {o_hrl, "./include"},
-                       {o_erl, "./src"}])
+                       {o_erl, "./src"},
+                       {i, "./src/proto"}])
        end || F <- string:tokens("$1", " ")],
        halt().
 endef
```

But a) this is just for the single concrete folder I'm using the test and not all folders that contain .proto files and
b) the import still doesn't seem to be working quite right for me afterwards, though that may be a gpb issue instead?

essen:
The file given as first argument also needs to be without the path. As for other folders, maybe just specify the option more than once for all the folders where proto files were found.

Bathtor:
> The file given as first argument also needs to be without the path. As for other folders, maybe just specify the option more than once for all the folders where proto files were found.

Right. I have no idea how to do either that^^
Maybe you could give it a try when you have some time? Would be much appreciated :)

essen:
I'll see if I have time before vacation starts, otherwise probably after.

PR #862, opened 2019-08-29
core/erlc.mk: Support non-US-ASCII characters in paths

dumbbell:
Without this patch, the added testcase would fail with the following error:

```
gmake[3]: *** No rule to make target '(...)/erlang.mk/test/h��test_core_makedep_non_usascii_paths/deps/test_core_makedep_non_usascii_paths_dep/include/hello.hrl', needed by 'src/hello.erl'.  Stop.
```

In this case, the path contains UTF-8-encoded `é` characters and they appear to be converted to ISO-8859-1 in the `makedep.erl` script.


dumbbell:
My initial patch was way too complicated. After re-reading the [Unicode usage](http://erlang.org/doc/apps/stdlib/unicode_usage.html) documentation, I understood that the conversion was done by the `file:write_file()` function. Converting the string just before passing it to the `file` module is enough to fix the problem.

essen:
Nice!

essen:
Merged, thanks! Push in a few.

PR #869, opened 2019-11-29
Add mysql-otp to the index

juhlig:
This PR removes the old `mysql` package and replaces it with `mysql-otp`.


essen:
Merged, push later today, thanks!

PR #871, opened 2019-12-09
plugins/c_src.mk: Deduce directory to create from $@

dumbbell:
... instead of hard-coding `priv`.


essen:
Merged, push later today, thanks!

PR #877, opened 2020-04-02
test/core_query.mk: Pin `farwest` in `core-query-deps-no-duplicates`

dumbbell:
It was previously "pinned" to its `master` branch. Unforunately, the testcase started to fail after `farwest` changed its own pinning of `cowlib`.

By pinning to a specific version (or commit in this case) of `farwest`, we make sure this won't happen again.


essen:
Merged, thanks!

PR #878, opened 2020-04-02
plugin_dialyzer.mk: Update erlang.mk in `deps/cowlib`

dumbbell:
The problem is that the copy of `erlang.mk` in cowlib does not have locking in `dep_autopatch_fetch_rebar`. This triggers failures in parallel testing of Erlang.mk.

A second improvement is to just test-build cowlib instead of actually running the tests. Test deps are fetched and compiled, but we don't waste time running cowlib's testsuite.

### DIFF
--- a/test/plugin_dialyzer.mk
+++ b/test/plugin_dialyzer.mk
@@ -282,8 +282,9 @@ dialyzer-plt-ebin-only: init
 	$i "Build the application"
 	$t $(MAKE) -C $(APP) $v
 
-	$i "Run Cowlib tests to fetch autopatched dependencies"
-	$t $(MAKE) -C $(APP)/deps/cowlib tests $v
+	$i "Build Cowlib for tests to fetch autopatched dependencies"
+	$t cp ../erlang.mk $(APP)/deps/cowlib
+	$t $(MAKE) -C $(APP)/deps/cowlib test-build $v
 
 	$i "Run Dialyzer"
 	$t $(DIALYZER_MUTEX) $(MAKE) -C $(APP) dialyze $v


essen:
I think the Cowlib Makefile should get patched by autopatch to have `include $(if $(ERLANG_MK_FILENAME),$(ERLANG_MK_FILENAME),erlang.mk)` and therefore it should have locking.

dumbbell:
Yes, it's even better to fix Cowlib. This is just the patch which unblocked me for now, feel free to go with the better solution :-)

essen:
No I mean the issue you see should not occur because Cowlib's erlang.mk file is not supposed to be used. Autopatch makes it so that the top-level erlang.mk file is used. Maybe you are seeing a different issue.

dumbbell:
I understand now. After checking, cowlib's `Makefile` is patched, but `$(ERLANG_MK_FILENAME)` is never passed to tests and sub-makes.

The following patch fixes the problem for the testcase that failed for me before:
```patch
diff --git a/Makefile b/Makefile
index b6b245c..02e9604 100644
--- a/Makefile
+++ b/Makefile
@@ -43,6 +43,8 @@ all:
 lint: all
        $(MAKE) -f erlang.mk --warn-undefined-variables

+check: export ERLANG_MK_FILENAME=$(realpath $(ERLANG_MK))
+
 ifdef p
 # Remove p from the list of variables since that conflicts with bootstrapping.
 MAKEOVERRIDES := $(filter-out p=$p,$(MAKEOVERRIDES))
```

I'm testing it with the entire testsuite and if it's ok, I will submit another pull request.

Closing this one.

essen:
Why wouldn't it be passed down? It's exported at the beginning of the erlang.mk file.

dumbbell:
I suppose because the top-level's project `Makefile` is bypassed:

```
dialyzer-plt-ebin-only: init
(...)
	$i "Run Cowlib tests to fetch autopatched dependencies"
	$t $(MAKE) -C $(APP)/deps/cowlib tests $v
```

essen:
Ah understood! No I'll just update Erlang.mk in Cowlib since the test depends on master. Please send a PR with your second fix though.

dumbbell:
Do you mean the patch above exporting `$(ERLANG_MK_FILENAME)`?

essen:
The change to use `test-build`, no point running tests as you said.

dumbbell:
Ok, submitted as #879.

PR #879, opened 2020-04-03
test/plugin_dialyzer.mk: Skip cowlib test in `dialyzer-plt-ebin-only`

dumbbell:
This is a waste of time and if they were to fail, this does not serve the Erlang.mk testsuite.


essen:
Merged, thanks!

PR #880, opened 2020-04-03
core/test.mk: Build application before its tests

dumbbell:
This fixes compiler warnings if modules under `tests` implement behaviors defined in `src`. This way, we make sure behaviors are compiled before the test modules.


essen:
Ah thanks, I've been eyeing that one for a long time. :-)

essen:
Merged, thanks!

Issue #822, opened 2019-01-24
Don't touch all files to recompile everything


essen:
It should be possible to have a single temporary file as a dependency to all others and use that to force the rebuild of everything, rather than touch everything. Perhaps the .app file can be used for this? If it's touched to 000 it should rebuild all of it. To be experimented with.

crownedgrouse:
Note that a hrl file may include other ones. Recreate a reverse tree with soft links could be an idea but maybe hard to be done on windows

essen:
Not sure what you mean or how it's related.

crownedgrouse:
let say   mod1.erl  includes inc1.hrl  which is including inc11.hrl and inc12.hrl.   mod2.erl includes inc12.hrl.
Editing inc12.hrl needs to recompile mod1.erl and mod2.erl . 

`.
├── .erlang.mk
│   └── includes
│       ├── include
│       │   ├── inc11.hrl
│       │   │   └── inc1.hrl -> ../../src/inc1.hrl
│       │   └── inc12.hrl
│       │       ├── inc1.hrl -> ../../src/inc1.hrl
│       │       └── mod2.erl -> ../../../../src/mod2.erl
│       └── src
│           └── inc1.hrl
│               └── mod1.erl -> ../../../../src/mod1.erl
├── include
│   ├── inc11.hrl
│   └── inc12.hrl
└── src
    ├── inc1.hrl
    ├── mod1.erl
    └── mod2.erl
`
finding files to touch when inc12.hrl is newer than last compilation would be 
`
$> find -L ` find .erlang.mk/  -name inc12.hrl -type d ` -type f 
.erlang.mk/includes/include/inc12.hrl/mod2.erl
.erlang.mk/includes/include/inc12.hrl/inc1.hrl/mod1.erl
`

crownedgrouse:
Editing a '.erl' file would need to update the tree first.

essen:
This ticket is about the "recompile everything when the Makefile changes" in case that's not clear. It currently has to touch all source files to accomplish that and that's not optimal. I'll edit the title.

crownedgrouse:
Indeed I did not understood the subject. BTW how is currently managed hrl files changes with erl files, in Erlang.mk? Not clear to me tgen

essen:
There's a makedep/depend step that generates a .d file with those dependencies.

essen:
[...]

PR #881, opened 2020-04-03
core/test.mk: Rebuild out-of-date test modules only

dumbbell:
... instead of always rebuilding all of them each time a testsuite is executed.

The only exception is when a Makefile was modified: like for main source files, test modules are all recompiled in this case.

### DIFF
--- a/core/test.mk
+++ b/core/test.mk
@@ -31,9 +31,23 @@ test-deps: $(ALL_TEST_DEPS_DIRS)
 endif
 
 ifneq ($(wildcard $(TEST_DIR)),)
-test-dir:
-	$(gen_verbose) erlc -v $(TEST_ERLC_OPTS) -o $(TEST_DIR) \
-		-pa ebin/ -I include/ $(call core_find,$(TEST_DIR)/,*.erl)
+test-dir: $(ERLANG_MK_TMP)/$(PROJECT).last-testdir-build
+	@:
+
+test_erlc_verbose_0 = @echo " ERLC  " $(filter-out $(patsubst %,%.erl,$(ERLC_EXCLUDE)),\
+	$(filter %.erl %.core,$(notdir $(FILES_TO_COMPILE))));
+test_erlc_verbose_2 = set -x;
+test_erlc_verbose = $(test_erlc_verbose_$(V))
+
+define compile_test_erl
+	$(test_erlc_verbose) erlc -v $(TEST_ERLC_OPTS) -o $(TEST_DIR) \
+		-pa ebin/ -I include/ $(1)
+endef
+
+ERL_TEST_FILES = $(call core_find,$(TEST_DIR)/,*.erl)
+$(ERLANG_MK_TMP)/$(PROJECT).last-testdir-build: $(ERL_TEST_FILES) $(MAKEFILE_LIST)
+	$(eval FILES_TO_COMPILE := $(if $(filter $(MAKEFILE_LIST),$?),$(filter $(ERL_TEST_FILES),$^),$?))
+	$(if $(strip $(FILES_TO_COMPILE)),$(call compile_test_erl,$(FILES_TO_COMPILE)); touch $@)
 endif
 
 test-build:: IS_TEST=1
@@ -70,5 +84,5 @@ clean:: clean-test-dir
 
 clean-test-dir:
 ifneq ($(wildcard $(TEST_DIR)/*.beam),)
-	$(gen_verbose) rm -f $(TEST_DIR)/*.beam
+	$(gen_verbose) rm -f $(TEST_DIR)/*.beam $(ERLANG_MK_TMP)/$(PROJECT).last-testdir-build
 endif
--- a/test/core_app.mk
+++ b/test/core_app.mk
@@ -1682,7 +1682,7 @@ core-app-yrl-header: init
 	$t mkdir $(APP)/
 	$t cp ../erlang.mk $(APP)/
 	$t $(MAKE) -C $(APP) -f erlang.mk bootstrap-lib $v
-	
+
 	$i "Create a .yrl file"
 	$t echo "Nonterminals E T F." > $(APP)/src/y_parse.yrl
 	$t echo "Terminals '+' '*' '(' ')' number." >> $(APP)/src/y_parse.yrl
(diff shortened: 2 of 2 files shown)


dumbbell:
In the context of rabbitmq-server, `time make test-build` (when there is nothing to do) goes from 7 seconds to 1.2 seconds.

essen:
I've been wondering about that as well, thanks for doing it!

core/test.mk
```diff
@@ -31,9 +31,20 @@ test-deps: $(ALL_TEST_DEPS_DIRS)
 endif
 
 ifneq ($(wildcard $(TEST_DIR)),)
-test-dir:
-	$(gen_verbose) erlc -v $(TEST_ERLC_OPTS) -o $(TEST_DIR) \
-		-pa ebin/ -I include/ $(call core_find,$(TEST_DIR)/,*.erl)
+test-dir: $(TEST_DIR)/.last-build
```
essen, line 34: We already have the file `ebin/test` so I don't think we should add more.

dumbbell: But `ebin/test` controls a different thing: whether the main modules were compiled for testing or not. If the application is recompiled for release meanwhile, `ebin/test` is removed and thus, all test modules will be recompiled if they depend on it, no matter if they are up-to-date or not.

Also, the fact that Erlang.mk checks for the existence of `src` and `test` independently indicates that a tests-only project is possible, in which case the `ebin` directory would not be recreated (and would not make sense just to hold `ebin/test`).

essen: OK. In any case we should probably use `.erlang.mk` for this if possible. (Same goes for `ebin/test` tbh.)

dumbbell: I made the change to use `.erlang.mk/last-testdir-build` as you suggested.

core/test.mk
```diff
@@ -31,9 +31,20 @@ test-deps: $(ALL_TEST_DEPS_DIRS)
...
 
 ifneq ($(wildcard $(TEST_DIR)),)
-test-dir:
-	$(gen_verbose) erlc -v $(TEST_ERLC_OPTS) -o $(TEST_DIR) \
-		-pa ebin/ -I include/ $(call core_find,$(TEST_DIR)/,*.erl)
+test-dir: $(ERLANG_MK_TMP)/last-testdir-build
+	@:
+
+ERL_TEST_FILES = $(call core_find,$(TEST_DIR)/,*.erl)
+$(ERLANG_MK_TMP)/last-testdir-build: $(ERL_TEST_FILES) $(MAKEFILE_LIST)
```
essen, line 38: Can't we have `$(ERL_TEST_FILES): $(MAKEFILE_LIST)` instead of the conditional logic? Or is that too likely to cause trouble?

essen:
Please also add a small test to confirm it works as intended.

core/test.mk
dumbbell: I don't see how this would work. Do you mean that I should add a recipe like the following one?

```
$(ERL_TEST_FILES): $(MAKEFILE_LIST)
        touch $(ERL_TEST_FILES)
```

dumbbell:
I added a testcase and pushed.

dumbbell:
There are several failures on CentOS in CI that I can't reproduce locally. What do you do usually to debug such errors?

essen:
Forgot to disable `-j` on the PR pipeline, CentOS is using an old Make or something and `-j` causes failures. Next push should succeed.

essen:
Windows is being annoying again, I'll fix it later today.

dumbbell:
Ok, I just force-pushed to trigger a new build.

essen:
Alpine is a flaky test that I'm tentatively improving in the most recent commit. Anyway don't bother rebuilding.

core/test.mk
essen: So I didn't look in details yet but if this method of yours is better can it be applied to the other source files? And then we get rid of the editor problem?

core/test.mk
```diff
@@ -31,9 +31,20 @@ test-deps: $(ALL_TEST_DEPS_DIRS)
...
 ifneq ($(wildcard $(TEST_DIR)),)
-test-dir:
-	$(gen_verbose) erlc -v $(TEST_ERLC_OPTS) -o $(TEST_DIR) \
-		-pa ebin/ -I include/ $(call core_find,$(TEST_DIR)/,*.erl)
+test-dir: $(ERLANG_MK_TMP)/$(PROJECT).last-testdir-build
+	@:
+
+ERL_TEST_FILES = $(call core_find,$(TEST_DIR)/,*.erl)
+$(ERLANG_MK_TMP)/$(PROJECT).last-testdir-build: $(ERL_TEST_FILES) $(MAKEFILE_LIST)
+	$(gen_verbose) set -e; \
```
essen, line 39: Hm `gen_verbose` probably prints something ugly, would prefer if we were printing ERL files to be compiled like when compiling normally. Since we don't recompile everything anymore.

core/test.mk
essen: This is what I meant but it probably doesn't work if not compiling the files one by one. https://github.com/ninenines/erlang.mk/blob/master/plugins/c_src.mk#L106

core/test.mk
dumbbell: Yeah, you are right, the produced output is uninformative. I will fix that.

dumbbell: I pushed the change. In the end, the recipe is simpler after using more GNU Make commands.

test/core_app.mk
```diff
@@ -2743,3 +2743,46 @@ endif
...
+	$t $(MAKE) -C $(APP) test-build $v
+
+	$i "Check that all compiled files exist"
+	$t test -f $(APP)/test/use_blue.beam
+	$t test -f $(APP)/test/use_red.beam
+
+# We will sleep for 1 second in several places, in case the underlying
+# filesystem does not support sub-second timestamps. This is to make
+# sure that we can compare file timestamps effectively.
+	$t sleep 1 && touch $(APP)/build-1
```
essen, line 2769: Please use `$t $(SLEEP)` for this, see `test/Makefile`, it'll only do `sleep 1` on systems that don't support sub-second timestamps.

essen: Same for the other `sleep 1`.

essen:
Just one minor comment about the test and we're good!

core/test.mk
dumbbell: Should we file a separate issue to look at the improvement you suggest?

test/core_app.mk
dumbbell: Nice! Doing the change now.

dumbbell: Done!

core/test.mk
essen: See #822.

essen:
Merging, thanks!

PR #882, opened 2020-04-09
esh.mk plugin

crownedgrouse:
Adding a new erlang.mk plugin allowing to generate files from template using shell.
Very usefull for config, vm.args (and any other) files depending of environment or OS.


crownedgrouse:
fixing some issues. closing

PR #883, opened 2020-04-11
Esh.mk plugin

crownedgrouse:
Add a new plugin for powerful, simple, esh template engine.

### DIFF
--- a/doc/src/guide/external_plugins_list.asciidoc
+++ b/doc/src/guide/external_plugins_list.asciidoc
@@ -26,6 +26,11 @@ language for the BEAM.
 An https://github.com/inaka/elvis.mk[Elvis plugin] for Erlang.mk.
 Elvis is an https://github.com/inaka/elvis[Erlang style reviewer].
 
+=== esh.mk
+
+An https://github.com/crownedgrouse/esh.mk[Esh template plugin] for Erlang.mk.
+https://github.com/jirutka/esh[Esh] is a simple template engine using shell.
+
 === geas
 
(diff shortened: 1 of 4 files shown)


crownedgrouse:
`make check p=esh_mk` failed due to .app file missing. Normal, it is only a erlang.mk plugin. Did not find a smart way to mute this error.
Could be smart to tag package that are only a make pluging  not an erlang project.

essen:
See https://github.com/ninenines/erlang.mk/blob/master/test/Makefile#L168

crownedgrouse:
I saw it, but error is raised before L199. This variable is used only L207

essen:
I mean sure if you call it explicitly but that doesn't matter, what matters is that it's not called when calling `make packages`.

crownedgrouse:
Ah ok. Do you mean I need to add this package in this variable anyway?

essen:
Yes please.

crownedgrouse:
https://github.com/ninenines/erlang.mk/blob/master/test/Makefile#L200 could replace 'false' by a test -f xxx/plugins.mk ?

crownedgrouse:
work for me with this fix .  `make check p=esh_mk` does not fails anymore.

essen:
I've merged locally and reverted the `test -f ...` lines. `make check p=...` is meant for Erlang applications. If for example Cowboy failed to build (no .app file) the error would not be detected because Cowboy also has a plugins.mk file. And if I remember correctly plugins are not required to have a `plugins.mk`, the plugin can be loaded from a different location. Will push soon, thanks!

PR #886, opened 2020-04-13
Do minimal checks to pkg_add.sh args

crownedgrouse:
Erlang.mk plugins are offen called with a `.mk` name, while it is not allowed to have dots in name for check tests. 
Do not create index file in such case , and ask to replace by an underscore.

### DIFF
--- a/pkg_add.sh
+++ b/pkg_add.sh
@@ -1,5 +1,18 @@
 #!/bin/sh
 
+if [ -z "$1$2$3$4$5$6" ]
+then
+	echo "Missing arguments" 1>&2
+	exit 1
+fi
+
+T=`echo "$1" | grep "\." | wc -l `
+if [ "$T" -eq "1" ] 
+then
+	echo "Invalid dot character in name. Considere replace by an underscore." 1>&2
+	exit 1
+fi 
+
 PKG_FILE="index/$1.mk"
 CONTENTS="PACKAGES += $1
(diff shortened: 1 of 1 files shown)


pkg_add.sh
```diff
@@ -1,5 +1,18 @@
 #!/bin/sh
 
+if [ -z "$1$2$3$4$5$6" ]
+then
+	echo "Missing arguments" 1>&2
+	exit 1
+fi
+
+T=`echo "$1" | grep "\." | wc -l `
+if [ "$T" -eq "1" ]
```
essen, line 10: Doesn't this only detect if there's one dot? So it won't catch `ci.erlang.mk` then?

crownedgrouse: indeed right. I will fix.

crownedgrouse: in fact no. wc return is 1 if there is at least one dot in name, otherwise zero.

crownedgrouse:
```
$ ./pkg_add.sh ci.erlang.mk
Invalid dot character in name. Considere replace by an underscore.
```

essen:
Merged, thanks. But note that the goal is to move away from the package index eventually. To what, I don't know, because there are no real alternatives for Erlang, and I don't have time to work on it.

crownedgrouse:
Currently working on a block chain package manager for fun. Note for later?

PR #892, opened 2020-06-22
core/test.mk: Abort build if a test file fails to compile

dumbbell:
In commit c55d0dcd6182983521d4ab34478cfe4b093edb12, the recipe was improved to only rebuild out-of-date test files (not everything as it was the case before).

However in the process, the exit status of the Erlang compiler was not checked anymore, leading to build failures in the test directory to be ignored. This patch fixes this bug.


essen:
Makes sense.

essen:
Merged, thanks!

dumbbell:
Thank you!

PR #893, opened 2020-06-23
plugins/concuerror.mk: Use $(MAKE) instead of hard-coding `make`

dumbbell:
This fixes the plugin when e.g. GNU Make is installed as `gmake` (and `make` is another incompatible implementation).


essen:
Schoolboy error.

dumbbell:
I can't reproduce the error with Concuerror in CI. The only error I got locally is the one fixed by this patch.

Do you have more details about the failure in CI?

essen:
Yes it's because CI is using OTP-23 and Concuerror doesn't work with it at this time. Don't worry about it.

dumbbell:
Ah ok, thank you :-)

essen:
Merged, thanks!

PR #895, opened 2020-06-26
Issue 894 (Darwin and NIFs)

### LINKED ISSUE #894: Can not generate NIF correctly on Mac OS X
Dear author

On Mac OS X, when there is a depends of project which will generate a NIF by using rebar3 and pc plugins. It can not generate a shared library, instead it generates an executable file.
And this phenomenon doesn't  happen on My FreeBSD and OpenBSD production machine.
So I think this may be a bug. And please refer this code below, where I think it's not correct.

https://github.com/ninenines/erlang.mk/blob/944c69634a4914684ce0d4eaf0b4f5950e06993a/core/deps.mk#L572

Thanks & Regards.

DavidAlphaFox:
When the platform is darwin, SharedFlag compares the extensions of filename
to determine whether to add "-shared" or not
Fixes #894

### DIFF
--- a/core/deps.mk
+++ b/core/deps.mk
@@ -549,6 +549,15 @@ define dep_autopatch_rebar.erl
 				false -> [];
 				{_, PortEnv0} -> FilterEnv(PortEnv0)
 			end,
+			SharedFlag = fun
+							([],_) -> "\n";
+							(Ext,darwin) ->
+								case lists:member(Ext,[".so",".dylib"]) of
+									true -> " -shared\n";
+									_ -> "\n"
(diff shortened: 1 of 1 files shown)


essen:
Please provide a test case to reproduce the issue.

DavidAlphaFox:
Please refer this repo https://github.com/DavidAlphaFox/rebar3_deps
I create a very simple application with common test framework to show this bugs.
Thanks & Regards.

essen:
I have simplified your patch, tested and merged. Will push later today. Thanks!

PR #900, opened 2020-08-11
fix rebar dependency specification

galdor:
Rebar3 emits a warning for "raw" git dependency versions, and expects a
tuple indicating what the version actually is (a branch, tag or ref).

This warning propagates to any software directly or indirectly using
Gun.

This syntax also works with rebar2
(https://github.com/rebar/rebar/blob/master/src/rebar_deps.erl#L608).

### DIFF
--- a/core/compat.mk
+++ b/core/compat.mk
@@ -23,7 +23,7 @@ define compat_rebar_config
 $(call comma_list,$(foreach d,$(DEPS),\
 	$(if $(filter hex,$(call dep_fetch,$d)),\
 		{$(call dep_name,$d)$(comma)"$(call dep_repo,$d)"},\
-		{$(call dep_name,$d)$(comma)".*"$(comma){git,"$(call dep_repo,$d)"$(comma)"$(call dep_commit,$d)"}})))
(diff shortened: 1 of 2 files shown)


essen:
There's errors https://buildkite.com/ninenines/erlang-dot-mk-prs/builds/571#f82429ad-13c2-4949-b0ca-24d31d05f43e/55-739

I can help tomorrow.

essen:
That's fine for commits but will this work fine for tags and branches too? Is there some expectations around differentiating ref/branch/tag in the rebar.config?

essen:
The warning is admittedly annoying, but based on https://github.com/erlang/rebar3/blob/c8811d120dcee5f4e51fd3e9778d5ba18cf54ec0/apps/rebar/src/rebar_git_resource.erl#L100-L109 I don't think we should use `{ref, _}` as it expects a sha in there.

In our case it is better to have the warning as updating the deps using rebar will not work as expected as indicated, so silencing the warning is not an improvement.

A better patch would be to use something like https://stackoverflow.com/questions/18222634/given-a-git-refname-can-i-detect-whether-its-a-hash-tag-or-branch to identify whether this is a branch, tag or commit and to use the right tuple for rebar.

essen:
I have opened https://github.com/ninenines/erlang.mk/pull/1010 with the better fix I mentioned earlier. Closing this, thanks!

PR #901, opened 2020-08-19
Update jsx master -> main branch repository

jchassoul:
We need this lite change to use jsx with erlang.mk again.


essen:
Merged, thanks!

PR #902, opened 2020-09-07
An attempt to optionally use the compile server in erlc.

mbj4668:
We use the env var ERLC_USE_SERVER to control the behaviour, since
erlc already uses that variable.

Although this seems to work fine for our project, I am not sure that
it works correctly with files generated from .yrl etc.

### DIFF
--- a/core/erlc.mk
+++ b/core/erlc.mk
@@ -89,14 +89,17 @@ define app_file
 endef
 endif
 
+ifneq ($(ERLC_USE_SERVER), true)
 app-build: ebin/$(PROJECT).app
 	$(verbose) :
+endif
 
 # Source files.
 
 ALL_SRC_FILES := $(sort $(call core_find,src/,*))
 
 ERL_FILES := $(filter %.erl,$(ALL_SRC_FILES))
+ORIG_ERL_FILES := $(ERL_FILES)
 CORE_FILES := $(filter %.core,$(ALL_SRC_FILES))
 
 # ASN.1 files.
@@ -328,6 +331,7 @@ ebin/%.beam: src/%.erl
 ERL_BEAM_FILES += $(patsubst src/%.erl, ebin/%.beam, $(ERL_FILES))
 
 app:: $(ERL_BEAM_FILES)
+app-build: ebin/$(PROJECT).app $(ERL_BEAM_FILES)
 endif
 
 ebin/$(PROJECT).app:: $(ERL_FILES) $(CORE_FILES) $(wildcard src/$(PROJECT).app.src)
--- a/core/erlc.mk
+++ b/core/erlc.mk
@@ -32,8 +32,13 @@ makedep_verbose_0 = @echo " DEPEND" $(PROJECT).d;
 makedep_verbose_2 = set -x;
 makedep_verbose = $(makedep_verbose_$(V))
 
+ifeq ($(ERLC_USE_SERVER), true)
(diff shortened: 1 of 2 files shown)


essen:
Yeah it definitely doesn't work for generated files.

mbj4668:
Yes.  I did some more fixes so that things like eunit works, but I think that you probably will do a better (correct) implementation anyway, so I think you can ignore this PR.  (That said, it works fine for our needs!)

essen:
It's tough to get right so any help is appreciated.

mbj4668:
Ok, I pushed the other changes we did.  In our project we have additional generated erl files (which we append to ERL_FILES), and it seems to work.

essen:
Could you provide the Makefile related to the generated files? Or at least the part directly related to that + where the `include erlang.mk` line is located with regard to that.

mbj4668:
Sure, here's the relevant snippet of the Makefile.

```
ERLC_USE_SERVER = true
export ERLC_USE_SERVER

EUNIT_OPTS = verbose,{print_depth,9999},{report,{eunit_surefire,[{dir,"."}]}}

include $(TOP_DIR)/mk/erlang.mk

ebin/$(PROJECT).app:: $(wildcard $(PROJECT_ENV_FILE))

# there is no variable LOCAL_BUILD_DEPS in erlang.mk, so we fake it here
ALL_DEPS_DIRS += $(addprefix $(LIB_DIR)/, $(LOCAL_BUILD_DEPS))

# ensure we have correct include path
ERLC_OPTS += -I $(LIB_DIR)
TEST_ERLC_OPTS += -I $(LIB_DIR)

# we allow export of variables from if/case
ERLC_OPTS := $(filter-out +warn_export_vars,$(ERLC_OPTS))
TEST_ERLC_OPTS := $(filter-out +warn_export_vars,$(TEST_ERLC_OPTS))

## add targets for yang compilation

YANG_FILES = $(filter %.yang,$(ALL_SRC_FILES))
YANG_ERL_FILES = $(patsubst src/%.yang, src/yy-%.erl, $(YANG_FILES))
ERL_FILES += $(YANG_ERL_FILES)
YANG_BEAM_FILES = $(patsubst src/%.yang, ebin/yy-%.beam, $(YANG_FILES))

app:: $(YANG_BEAM_FILES)

YANGER_OPTS ?=

YANGER_YY_DIR = $(LIB_DIR)/yanger_yy/ebin
YY_SRC = $(LIB_DIR)/yy/src
YY_DIR = $(LIB_DIR)/yy/ebin
SYST_SRC = $(LIB_DIR)/syst/src

yang_verbose_0 = @echo " YANGER " $(filter %.yang,$<);
yang_verbose_2 = set -x;
yang_verbose = $(yang_verbose_$(V))

src/yy-%.erl: src/%.yang ../yy/include/yy.hrl $(YANGER_YY_DIR)/yanger_yy.beam
	$(yang_verbose) $(YANGER) -Werror -P $(YANGER_YY_DIR) --pz $(YY_DIR) \
	-p $(SYST_SRC) -p $(YY_SRC) -f yy $(YANGER_OPTS) -o $@ $<

## if we have yang files we need yanger
ifneq ($(strip $(YANG_FILES)),)
LOCAL_BUILD_DEPS += yanger_yy
endif
```

essen:
Thanks! I'll take a look when time allows. Plan is to do this and package changes together as they're both breaking or potentially breaking changes.

mbj4668:
Ok.   +1 to the package changes as well!

PR #906, opened 2020-10-25
Add COVER_EXCL_MODS variable

josephDunne:
I added support for excluding modules from the cover report through a variable called COVER_EXCL_MODS.
Added tests and updated documentation.


josephDunne:
My tests pass locally. Something else is wrong with the build checks not related to this change

essen:
Merged, thanks!

PR #911, opened 2020-10-30
docs(edoc): recursively find source paths

pinicarus:
Subdirectories of `src` are correctly used when compiling, so they should also be considered when generating `edoc`.

### DIFF
--- a/plugins/edoc.mk
+++ b/plugins/edoc.mk
@@ -11,7 +11,11 @@ EDOC_OUTPUT ?= doc
 
 define edoc.erl
 	SrcPaths = lists:foldl(fun(P, Acc) ->
-		filelib:wildcard(atom_to_list(P) ++ "/{src,c_src}") ++ Acc
+		filelib:wildcard(atom_to_list(P) ++ "/{src,c_src}")
+		++ lists:filter(fun(D) ->
+			filelib:is_dir(D)
+		end, filelib:wildcard(atom_to_list(P) ++ "/{src,c_src}/**"))
+		++ Acc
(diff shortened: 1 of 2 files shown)


essen:
Thanks! Please add a test in test/plugin_edoc.mk.

pinicarus:
> Thanks! Please add a test in test/plugin_edoc.mk.

Thanks for your quick feedback.
I added two tests, but it seems the CI fails for some other reasons.

Also, this change only triggers when `EDOC_SRC_DIRS` is set.
Shouldn't there be some default for this variable ? Maybe `$(CURDIR)` ?

essen:
You need either the variable set or a doc/overview.edoc file. It's by design, not all projects that have source files or a doc/ directory are using edoc. There has to be a way to identify users of edoc and that's the current way to do it.

Don't worry about CI in general, there's a couple broken tests that I fixed in a branch and there might be other issues I didn't get to yet. Just make sure the tests pass for you and that'll be fine.

pinicarus:
Both tests (`edoc-src{,-multiapp}-subdirs`) are OK on all Linux distributions.

pinicarus:
Hi,

Do I need to change something else or do you think this PR is OK for merging ?

Thanks a lot.

essen:
In theory it's fine, I'll get to it as part of the changes I'm making currently. Might be a few weeks because I'm only doing a day/week on Erlang.mk at the moment and the tasks at hand are large.

pinicarus:
OK, thanks for your feedback.

essen:
Merged, thanks! Took me a while to get to but there were no issues at all. Will push later today. Cheers.

PR #913, opened 2020-11-12
Replace erlang:get_stacktrace() with catch C:E:S

zuiderkwast:
`erlang:get_stacktrace()` will be removed in OTP 24.

### DIFF
--- a/plugins/asciidoc.mk
+++ b/plugins/asciidoc.mk
@@ -57,8 +57,8 @@ try
 		})
 	end || F <- [$(shell echo $(addprefix $(comma)\",$(addsuffix \",$1)) | sed 's/^.//')]],
 	halt(0)
-catch C:E ->
-	io:format("Exception ~p:~p~nStacktrace: ~p~n", [C, E, erlang:get_stacktrace()]),
+catch C:E$(if $V,:S) ->
+	io:format("Exception: ~p:~p~n$(if $V,Stacktrace: ~p~n)", [C, E$(if $V,$(comma) S)]),
 	halt(1)
 end.
 endef
--- a/plugins/proper.mk
+++ b/plugins/proper.mk
@@ -37,8 +37,8 @@ define proper_check.erl
 	end of
 		true -> halt(0);
 		_ -> halt(1)
-	catch error:undef ->
-		io:format("Undefined property or module?~n~p~n", [erlang:get_stacktrace()]),
+	catch error:undef$(if $V,:Stacktrace) ->
+		io:format("Undefined property or module?~n$(if $V,~p~n)", [$(if $V,Stacktrace)]),
 		halt(0)
 	end.
 endef
--- a/plugins/triq.mk
+++ b/plugins/triq.mk
@@ -26,8 +26,8 @@ define triq_check.erl
 	end of
 		true -> halt(0);
 		_ -> halt(1)
-	catch error:undef ->
-		io:format("Undefined property or module?~n~p~n", [erlang:get_stacktrace()]),
+	catch error:undef$(if $V,:Stacktrace) ->
+		io:format("Undefined property or module?~n$(if $V,~p~n)", [$(if $V,Stacktrace)]),
 		halt(0)
 	end.
 endef


essen:
Thanks. I'll merge it as part of the breaking changes I'm currently working on.

crownedgrouse:
Be careful, this syntax is not compatible with older OTP release.
This should use some Macro magic testing version

essen:
Yes this will make Erlang.mk compatible with OTP-21.0+ only. I believe this is fine as people can stay on older versions if necessary. There will be a final tag before the breaking changes.

crownedgrouse:
Sure but is make erlang-mk aware of this?
This should take care of pinned otp version or even of the current erl version found. Otherwise an upgrade will break the makr

essen:
It will be mentioned in the CHANGELOG printed when you do `make erlang-mk`.

crownedgrouse:
Yeah, but too late MHO. How user can revert?

essen:
Err, `git checkout erlang.mk` or equivalent? We can easily provide instructions.

crownedgrouse:
I would really prefere something transparent and portable, using former syntaxe before 21 and new if get_stacktrace missing in exported functions.
But you are owner.

essen:
This is `erl -eval` we're talking about. So unless you have a solution for supporting both old and new OTP syntax in `erl -eval` I'm not sure what we can do about this. Note that it's always the top-level Erlang.mk that's being used so as long as your Erlang.mk is correct for your OTP version it's not going to create issues for you anyway. The few people that need to support old versions can temporarily stay on the older Erlang.mk. I don't think it's going to be too many people though. Even RabbitMQ is > OTP 21.0 by now.

zuiderkwast:
An alternative is to just `catch C:E`. These catches seem to be for debugging proper, triq and asciidoc. I don't think the stacktraces actually need to be there.

FYI, https://github.com/mysql-otp/mysql-otp is still compatible with OTP as old as R16B (released 2013-02-25), but that's mostly because it was possible to keep compatibility without much effort. There we're using a stripped-down erlang.mk with only a bare minimum such as eunit and dialyzer. We don't even have the deps plugin.

essen:
If I recall the stacktrace can be really important for proper and/or triq. And asciideck is a bit rough on the edges so it's currently important, but shouldn't be ultimately. That said perhaps the stacktrace can be shown if `V=1` (or >1) is set only.

essen:
Another point: `asciideck` is OTP-20+ right now but might see it's requirement increased. `proper` is currently OTP-19+. Didn't check `triq`. So whatever the evals do should depend on the relevant application's OTP requirements.

zuiderkwast:
> perhaps the stacktrace can be shown if V=1 (or >1) is set only.

This means conditional stuff in the Makefile.... `$(call ...)` inside Erlang code?

essen:
`$(if $(V),:Stacktrace) ... $(if $(V),$(comma) Stacktrace`

zuiderkwast:
Sure, that would work, but we'd get some crazy hard-to-read shit...

```
	catch error:undef:$(if $(V),:S) ->
		io:format("Undefined property or module?~n$(if $(V),~p~n)", [$(if $(V),S)]),
```

Perhaps it's better to break things out:

```
ifeq ($(V),1)
catch_undef=catch error:undef:S -> io:format("Undefined property or module?~n~p~n, [S]), halt(0)
else
catch_undef=catch error:undef -> io:format("Undefined property or module? Use V=1 to debug.~n", []), halt(0)
endif
```

and then use it in Erlang code

```
    try
        ...
    $(catch_undef)
    end
```

Either way, it's doable.

essen:
Dunno, doesn't look crazy unreadable to me. Things like this are done in other places. Can even do `$V` instead of `$(V)` to save a few characters.

zuiderkwast:
True, it's not more unreadable than the rest of erlang.mk. :-)

artman41:
unsure whether a decision was already made but I've used this syntax when doing changes to erlang.mk to support both erlang versions.

Don't know whether this is a clear enough solution to handle both Stacktrace variants

```Makefile
open_parenthesis := (
close_parenthesis := )
...
CAN_USE_NEW_TRYCATCH = $(if $(shell erl -noshell -eval 'try throw(ok) catch _:_:_ -> ok end, halt(0).' 2>/dev/null),,1)
...
catch E:R$(if $(CAN_USE_NEW_TRYCATCH),:S) ->
    $(if $(CAN_USE_NEW_TRYCATCH),ok,S = erlang:get_stacktrace$(open_parenthesis)$(close_parenthesis)),
```

essen:
Now that a few years have gone (we are (almost) at OTP-26), and considering this will not break the default usage, only when V=1 is used, I will merge this. Thanks!

PR #914, opened 2020-11-22
fix: replace perl usage with coreutils

pinicarus:
This should replace the dependency to perl

### DIFF
--- a/core/core.mk
+++ b/core/core.mk
@@ -191,8 +191,7 @@ core_lc = $(subst A,a,$(subst B,b,$(subst C,c,$(subst D,d,$(subst E,e,$(subst F,
 
 core_ls = $(filter-out $(1),$(shell echo $(1)))
 
-# @todo Use a solution that does not require using perl.
-core_relpath = $(shell perl -e 'use File::Spec; print File::Spec->abs2rel(@ARGV) . "\n"' $1 $2)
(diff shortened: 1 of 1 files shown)


essen:
Sounds good. Thanks!

essen:
Merged, thanks!

essen:
I have to revert this commit because `realpath` is not available on macOS by default.

pinicarus:
> I have to revert this commit because `realpath` is not available on macOS by default.

Oh, too bad.

essen:
Would be open to another (reasonable) solution of course.

crownedgrouse:
"readlink - f " can be used when realpath is missing

pinicarus:
> "readlink - f " can be used when realpath is missing

Yes, but it lacks the `--relative-to` feature.

pinicarus:
I have an alternative implementation that uses make's `abspath` (or `realpath`) function and an awk script that only needs the `split` function.

Running it on my laptop with the `--posix` flag gives the same output as `realpath`.
Do you known it `awk` is available in macOS by default?

essen:
Not sure about "by default" but the problem with awk is there's an alternative implementation that might not be compatible, used in Busybox distros for example. That said, GNU awk is required to even build Erlang.mk and the only problem was Alpine (where it's listed as a mandatory dependency in the user guide) so it's probably safe to use.

PR #915, opened 2020-11-26
ignore files with '#' in their names, e.g. emacs lock files

mbj4668:
If an erlang file has a # in its name (e.g. an emacs lock file), and it has dependencies, it will end up in the .d file with the #, which is is interpreted as a comment.

### DIFF
--- a/core/core.mk
+++ b/core/core.mk
@@ -184,8 +184,8 @@ core_http_get = curl -Lf$(if $(filter-out 0,$(V)),,s)o $(call core_native_path,$
 
 core_eq = $(and $(findstring $(1),$(2)),$(findstring $(2),$(1)))
 
-# We skip files that contain spaces because they end up causing issues.
-core_find = $(if $(wildcard $1),$(shell find $(1:%/=%) \( -type l -o -type f \) -name $(subst *,\*,$2) | grep -v " "))
+# We skip files that contain spaces or '#' because they end up causing issues.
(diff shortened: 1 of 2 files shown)


essen:
Thanks! Could you please add a test to `test/core_makedep.mk`?

essen:
Just to be clear this is the only place (makedep) where the issue pops up?

I'm not convinced about the fix. I think perhaps the pattern should just be better specified.

mbj4668:
This is the only place that I noticed this problem (specifically that the .d file was invalid)

There might be other characters except "#" and " " that become problematic, if that's what you mean with "better specified".

essen:
I mean that instead of `$(call core_find,src/,*)` it might be better to do something like `$(call core_find,src/,[^.]*)` aka all files that don't begin with a dot. I don't know if just replacing this would work, haven't tested it, but you get the idea.

mbj4668:
That probably works for this particular case.  But do see any issues with my solution?

essen:
For # it's probably possible to escape it so I'd rather see that. The same is not completely possible with spaces because spaces are used for separators.

Otherwise I'd rather have zero greps there if we can. Or at most one if not possible otherwise. I don't want to keep adding greps basically.

mbj4668:
My initial idea was to escape #, but I couldn't figure out how to do that.

I agree about the adding of greps.  An alternative could be to extend the current find pattern e.g. `-a -not -name "* #*`  and remove the greps.

essen:
Yes extending find sounds good.

essen:
Can you please rebase this against current master so tests that were fixed no longer fail in CI?

PR #918, opened 2020-11-30
Add sleep to test cases, if needed

mbj4668:
Try do detect if we need to sleep in tests between a build and when a
source file is modified.  Without a sleep the source file and the
target may get the same modification time, on some machines.


essen:
Merged, thanks!

PR #921, opened 2020-12-04
fix: replace perl usage with awk

pinicarus:
A new proposal, using `awk` this time


essen:
Hm yeah I'd rather have the Perl dependency over this as I'm sure you can understand. :-)

pinicarus:
> Hm yeah I'd rather have the Perl dependency over this as I'm sure you can understand. :-)

It's your call :)

PR #926, opened 2021-03-26
test/core_deps.mk: Fix jquery branch name

dumbbell:
Apparently it was renamed.


essen:
Both commits were in the other PR.

PR #927, opened 2021-03-26
Support rebar 1.2.0 lock files

dumbbell:


dumbbell:
The testsuite will likely fail. A fix for the failure was submitted separately:  #926.

lhoguin:
Thanks. I will do that on Monday, I don't trust me to do it on a Friday evening. :-)

essen:
I've merged this. I've fixed one more test (related to another master->main, in Cowlib this time, used in a test) and there's one test remaining that fails but that's because of Gun depending on Cowlib master, I will leave that for another time. Thanks!

PR #932, opened 2021-06-09
Introduce LOCAL_BUILD_DEPS and LOCAL_TEST_DEPS

pjk25:
`LOCAL_BUILD_DEPS` is the equivalent of `BUILD_DEPS` for local deps in `APPS_DIR`

`LOCAL_TEST_DEPS` is the equivalent of `TEST_DEPS` for local deps in `APPS_DIR`

Something that I noticed which working on this is that `APPS_DIR` and `DEPS_DIR` always end up in the `ERL_LIBS`. So, these values don't affect what's on the code path when tests run. They simply imply a compilation order.


essen:
cc @lhoguin

lhoguin:
OK in principle but I'd like to have a test for each or see where you have been using it first.

pjk25:
Sure, I did some manual testing but am happy to add something automated.

pjk25:
After reverting my changes and writing a test case for `LOCAL_BUILD_DEPS`, it seems that it’s actually unnecessary. The projects in `APPS_DIR` seem to be built automatically, and are always part of the `ERL_LIBS`. I will close this.

PR #936, opened 2021-10-06
Added static compilation option for NIFs

dominicletz:


plugins/c_src.mk
```diff
@@ -53,6 +64,11 @@ ifneq ($(PLATFORM),msys2)
 	CXXFLAGS += -fPIC
 endif
 
+ifeq (,$(STATIC_ERLANG_NIF))
```
dominicletz, line 67: Maybe this should be `ifeq ($(C_SRC_TYPE),static)` for clarity

essen:
Merged and test added in a separate commit. Push later today. Feedback more than welcome! Since it's a static build I only confirm it works as intended by calling `objdump`, but I haven't built a full OTP with the static NIF. Cheers.

PR #937, opened 2021-10-19
Removed hardcoded `-arch x86_64` for MacOS nifs

dominicletz:
Just checked git blame and this flag seems to have been there from the very beginning (powerpc remenant?). For current macs the flag seems problematic as it overrides the default choice of building for the current architecture and hence prevents building on ARM based macs.

### DIFF
--- a/plugins/c_src.mk
+++ b/plugins/c_src.mk
@@ -35,9 +35,9 @@ ifeq ($(PLATFORM),msys2)
 	CXXFLAGS ?= -O3 -finline-functions -Wall
 else ifeq ($(PLATFORM),darwin)
 	CC ?= cc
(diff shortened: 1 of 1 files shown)


dominicletz:
Another thought on this: Even more convenient might be a check for the `arch` of the OTP system in `$PATH` to ensure the resulting NIFs are always compatible with that. E.g. with Rosetta2 an ARM user can install and use an intel based OTP build, and might even use that for building. -- Not sure though atm how to best get the right arch flags from `erl`...

essen:
I will get to this when I am back at work. @lhoguin

essen:
Merged, thanks!

PR #952, opened 2022-06-10
Fix compatibility with Relx v4 regarding {git, short} && {git, long} versions

spylik:
Relx v4 supports keys `{git, short}` and `{git, long}` for determine version based on git revision.

related `relx` routine is there: https://github.com/erlware/relx/blob/c94a1d016a26109a9f6e482d0cf0a3418cc60fa2/src/rlx_config.erl#L188

so, this PR is just adding similar routine to erlang.mk for support new `relx.config`s


essen:
Merged, push later today. Thanks!

PR #958, opened 2022-08-16
Macs are not all x86 anymore

sstrollo:
Seems strange to have a hard-coded arch when building C programs on Darwin. Suggest they are simply removed, as in this pull request.


sstrollo:
Oops, just noticed that there already is a pull request for this: https://github.com/ninenines/erlang.mk/pull/937

essen:
It's my bad I didn't take it on soon enough (that and I can't test on the new Macs). I'm happy to merge either of them, good to have confirmation that this is what's needed. I will get to it when I am back at work. Thanks!

essen:
Merging the other one. Thanks!

sstrollo:
Great - thanks @essen !

PR #967, opened 2023-01-05
Add Common Test Sub Group example

ansd:
> With this type of specification execution of unwanted groups (in otherwise matching paths), and/or the execution of subgroups can be avoided. The command line syntax of the group path is a list of group names in the path

https://www.erlang.org/doc/apps/common_test/run_test_chapter.html#test-case-group-execution


essen:
Merged, thanks!

PR #968, opened 2023-01-29
Allow git + hex deps to be cached

artman41:
- Enabled via `CACHE_DEPS=1`
- Cached files are beneath `$(HOME)/.cache/erlang.mk/` by default
- Will cache Git Fetches (including submodules) & Hex fetches
  - These cache in `$(CACHE_DIR)/git` & `$(CACHE_DIR)/hex`

### DIFF
--- a/core/deps.mk
+++ b/core/deps.mk
@@ -687,13 +687,15 @@ define dep_autopatch_appsrc.erl
 	halt()
 endef
 
+ifeq ($(CACHE_DEPS),1)
+
 define __fetch_git
 	mkdir -p $(CACHE_DIR)/gits; \
 	if test -d "$(join $(CACHE_DIR)/gits/,$(call dep_name,$(1)))"; then \
 		cd $(join $(CACHE_DIR)/gits/,$(call dep_name,$(1))); \
 		if ! git checkout -q $(call dep_commit,$(1)); then \
 			git remote set-url origin $(call dep_repo,$(1)) && \
-			git fetch && \
+			git fetch && git pull && \
 			git checkout -q $(call dep_commit,$(1)); \
 		fi; \
 	else \
@@ -701,8 +703,6 @@ define __fetch_git
 	fi; \
 	git clone -q --branch --single-branch $(call dep_commit,$(1)) -- $(join $(CACHE_DIR)/gits/,$(call dep_name,$(1))) $(2)
 endef
-
-ifeq ($(CACHE_DEPS),1)
 
 define dep_fetch_git
 	$(call __fetch_git,$(1),$(DEPS_DIR)/$(call dep_name,$(1)));
--- a/core/deps.mk
+++ b/core/deps.mk
@@ -24,8 +24,10 @@ export REBAR_DEPS_DIR
 REBAR_GIT ?= https://github.com/rebar/rebar
 REBAR_COMMIT ?= 576e12171ab8d69b048b827b92aa65d067deea01
 
-CACHE_DIR ?= $(HOME)/.cache/erlang.mk
 CACHE_DEPS ?= 0
+
+CACHE_DIR ?= $(if $(XDG_CACHE_HOME),$(XDG_CACHE_HOME),$(HOME)/.cache)/erlang.mk
+export CACHE_DIR
 
 # External "early" plugins (see core/plugins.mk for regular plugins).
 # They both use the core_dep_plugin macro.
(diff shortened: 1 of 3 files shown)


artman41:
The reasoning behind this MR is that, recently, I've been doing a lot of offline dev & it's quite frustrating when you know you have the git repo in a project _somewhere_ locally, but not sure where.

Aside from offline benefits, this MR should also speed up subsequent `deps` calls after a `distclean` if you're on a slow connection & using the cache.

essen:
Not a bad idea but I think you are missing a `git pull` when the repository already exists otherwise it might not be able to checkout new branches/tags?

artman41:
> Not a bad idea but I think you are missing a `git pull` when the repository already exists otherwise it might not be able to checkout new branches/tags?

Good shout, threw a `git pull` in there after the `git fetch` - anything else you can think of?

Unsure whether it's worthwhile symlinking from the `$(CACHE_DIR)` or whether using it as a clone source is good enough

essen:
A clone would work out better across projects, for example if one project depends on a version and another a different version. Which means that the git checkout after the pull is probably not necessary. Also git pull includes git fetch so that's not necessary either, one or the other should be good enough. Not sure if you had git fetch before, maybe I didn't look hard enough.

artman41:
> A clone would work out better across projects, for example if one project depends on a version and another a different version. Which means that the git checkout after the pull is probably not necessary. Also git pull includes git fetch so that's not necessary either, one or the other should be good enough. Not sure if you had git fetch before, maybe I didn't look hard enough.
> 
> Other than that, I think having some tests would be good. I'll see if some folks want to try it tomorrow.

Sounds good, I've had it in the past where a pull only pulled the latest refs for the branches I had locally (though I guess if we change it to `git pull --all` then it'll likely pull everything)

- I Agree with the cloning over symlinking.

Ref the `git checkout` - you're probably right that we can remove that, now we're using it as a locally git repo, leaving it checked out on to a branch doesn't really have a purpose for us.

What're you thinking test wise?

My current idea is the following (otherwise I'm not sure how you'd be able to tell you've cloned it from the cache only)

- Check dep doesn't exist in cache
- Clone dep in
- Check dep does exist in cache
- _Make a commit to the cache version only_
- distclean
- _Clone deps in again, check cache commit is there_

I also had a realisation while typing this that it'd likely be worthwhile to have a target to clear the cache (maybe with the ability to specify a specific dep?) - what're your thoughts on that?

essen:
The tests have code you can reuse that creates repos. I would not recommend modifying the cache within the test since we want to test the cache as well not just the code. There's probably a bunch of different scenarios to try, but separate git/subfolder/hex would be a good start. Also testing commits, branches and tags and making sure different projects can depend on different versions and they all can use the same cache.

core/deps.mk
```diff
@@ -24,6 +24,9 @@ export REBAR_DEPS_DIR
 REBAR_GIT ?= https://github.com/rebar/rebar
 REBAR_COMMIT ?= 576e12171ab8d69b048b827b92aa65d067deea01
 
+CACHE_DIR ?= $(HOME)/.cache/erlang.mk
```
dumbbell, line 27: For a better integration with faily recent Linux/Unix systems, it would be nice to follow the [XDG user directories specification](https://specifications.freedesktop.org/basedir-spec/basedir-spec-latest.html):

>  $XDG_CACHE_HOME defines the base directory relative to which user-specific non-essential data files should be stored. If $XDG_CACHE_HOME is either not set or empty, a default equal to $HOME/.cache should be used.

dumbbell, line 27: For a better integration with faily recent Linux/Unix systems, it would be nice to follow the [XDG user directories specification](https://specifications.freedesktop.org/basedir-spec/basedir-spec-latest.html):

>  $XDG_CACHE_HOME defines the base directory relative to which user-specific non-essential data files should be stored. If $XDG_CACHE_HOME is either not set or empty, a default equal to $HOME/.cache should be used.

artman41: good idea, I've made the change 👍

artman41:
> The tests have code you can reuse that creates repos. I would not recommend modifying the cache within the test since we want to test the cache as well not just the code. There's probably a bunch of different scenarios to try, but separate git/subfolder/hex would be a good start. Also testing commits, branches and tags and making sure different projects can depend on different versions and they all can use the same cache.
> 
> About clearing the cache, nothing beats rm -rf ~/.cache/erlang.mk/...

- New target `cacheclean`
- 2 Tests, 1 to test that the cache works & 1 to test that 2 apps can reuse the cache for different versions of the same dep

PR #972, opened 2023-04-06
Update references to rebar3

mkuratczyk:
* Switch to the current rebar repo
* Use a recent version of rebar


essen:
I will continue the work on this later this year.

essen:
I am about to push changes that did this and more.

Issue #976, opened 2023-05-19
OTP26 dialyzer support: `--no_native` flag is gone


mkuratczyk:
The `--no_native` flag in `dialyzer` was ignored since OTP 24.0 and [got removed](https://github.com/erlang/otp/commit/0190550ecfe1bb03370b9c711ddab9b5f4afa625#diff-58be20d83a1e7dae76a736be0938f3596375ed8a739a6390da4db46fbe234c18L76) between 26.0-rc3 and 26.0 final.

Therefore `make dialyze` on 26.0 fails with `dialyzer: No such file, directory or application: "no_native"`.

Simply removing this option makes it work on OTP26 but I guess could have some impact on OTP23 and older.

essen:
Please open a PR removing it the recent changes I have done have removed a bunch of other native related code anyway. Missed this one I guess.

PR #977, opened 2023-05-19
Remove deprecated --no_native option from dialyzer

mkuratczyk:
Fixes https://github.com/ninenines/erlang.mk/issues/976


essen:
Merged, thanks!

PR #979, opened 2023-06-04
Add Elixir.mk - Allows for use of Elixir deps within an Erlang project.

artman41:
This change introduces a new core plugin called `elixir.mk`.

This change accomplishes the following:
- Converts `mix.exs` projects to use `erlang.mk`
- Compiles `.ex` files, allowing for a project to use either Erlang, Elixir or a combination of both
- Properly configures paths so that Elixir dependencies are added to the `$(PROJECT).app`
- Treats Elixir as a dependency & fetches it as such, rather than relying on a system installation

This change makes use of `hex.mk`


artman41:
Currently rebasing the change :)

artman41:
@essen I've added a test under `test/core_elixir.mk` though I've been testing locally with a sample project.

essen:
Hello! I've not had a good look yet but I doubt this works with some Elixir NIFs since they often have a Makefile and thus wouldn't enter the elixir autopatch.

artman41:
> Hello! I've not had a good look yet but I doubt this works with some Elixir NIFs since they often have a Makefile and thus wouldn't enter the elixir autopatch.

Damn, that's a good shout - any idea of an elixir lib which uses NIFs

essen:
There's a few in this search but gotta filter. Oftentimes if the page says mix + make as a build tool, that's an Elixir NIF. But not always.

https://hex.pm/packages?search=nif&sort=total_downloads

artman41:
> There's a few in this search but gotta filter. Oftentimes if the page says mix + make as a build tool, that's an Elixir NIF. But not always.
> 
> https://hex.pm/packages?search=nif&sort=total_downloads

I've just commited  a change which _should_ now make it support Elixir Nifs (assuming they use elixir_make - I'm not 100% familiar with the Elixir NIF ecosystem so not quite sure if I've accounted for the right things there).

I've added an additional test which will attempt to compile `libsalty2`, though this depends on libsodium.

If you've any more suggestions @essen, I'm more than happy to add them 👍

artman41:
Any chance I could poke you to look this over @essen? 😄

core/elixir.mk
```diff
@@ -0,0 +1,314 @@
...
+	$(if $(PROJECT_MOD),{mod$(comma) {'$(PROJECT_MOD)'$(comma) []}}$(comma),)
+	{env, $(subst \,\\,$(PROJECT_ENV))}$(if $(findstring {,$(PROJECT_APP_EXTRA_KEYS)),$(comma)$(newline)$(tab)$(subst \,\\,$(PROJECT_APP_EXTRA_KEYS)),)
+]}.
+endef
+
+app:: $(if $(wildcard ebin/test),clean) deps
+	$(verbose) $(MAKE) --no-print-directory $(PROJECT).d
+	$(verbose) $(MAKE) --no-print-directory app-build
+
+app-build: ebin/$(PROJECT).app
```
essen, line 43: Should be `app-build::`.

essen:
Does this work even if using local system installed Elixir?

artman41:
> Does this work even if using local system installed Elixir?

In what sense @essen?

I'm expecting elixir to always be loaded as a dep of the project rather than as an installed version on the system - I can try and account for this though if you'd prefer that functionality to be in there.

I'll edit this for confirmation but the only thing I think that could be problematic with an installed version of elixir is if `code:get_path()` returned paths to the installed elixir version, though I wouldn't ever expect that to be the case with a standard install.

essen:
I think we should take the dep Elixir if one is available as a dep, and the system Elixir otherwise. Basically the dep can be used as an override.

Most projects don't require a specific version of Elixir, and for example in RabbitMQ we explicitly use the system Elixir (we even build packages for it).

artman41:
@essen could you have a look over when you have chance?

Based on your feedback, I've updated the MR to default to using the System Install of Elixir if available. This can be manually overriden by a user-set var though (`ELIXIR_USE_SYSTEM`)

Let me know what you think :)

artman41:
> @essen could you have a look over when you have chance?
> 
> Based on your feedback, I've updated the MR to default to using the System Install of Elixir if available. This can be manually overriden by a user-set var though (`ELIXIR_USE_SYSTEM`)
> 
> Let me know what you think :)

just poking you 😄

essen:
I'll get to it no need to remind so often.

artman41:
@essen sorry to poke on this one again, any idea when you'll be able to get round to it? :)

core/deps.mk
```diff
@@ -148,7 +148,7 @@ endif
 
 # Core targets.
 
-ALL_APPS_DIRS_TO_BUILD = $(if $(LOCAL_DEPS_DIRS)$(IS_APP),$(LOCAL_DEPS_DIRS),$(ALL_APPS_DIRS))
+ALL_APPS_DIRS_TO_BUILD = $(if $(filter-out $(ELIXIR_BUILTINS),$(LOCAL_DEPS_DIRS))$(IS_APP),$(filter-out $(ELIXIR_BUILTINS),$(LOCAL_DEPS_DIRS)),$(ALL_APPS_DIRS))
```
essen, line 151: I don't think Elixir builtins should be in `LOCAL_DEPS_DIRS` to begin with. It's defined as:

```
LOCAL_DEPS_DIRS = $(foreach a,$(LOCAL_DEPS),$(if $(wildcard $(APPS_DIR)/$(a)),$(APPS_DIR)/$(a)))
```

It's not meant to contain applications the user didn't write.

core/elixir.mk
```diff
@@ -0,0 +1,334 @@
...
+	{description, "$(PROJECT_DESCRIPTION)"},
+	{vsn, "$(PROJECT_VERSION)"},$(if $(IS_DEP),
+	{id$(comma)$(space)"$(1)"}$(comma))
+	{modules, [$(call comma_list,$(2))]},
+	{registered, [$(call comma_list,$(PROJECT)_sup $(PROJECT_REGISTERED))]},
+	{applications, [$(call comma_list,kernel stdlib $(OTP_DEPS) $(LOCAL_DEPS) $(foreach dep,$(DEPS),$(call dep_name,$(dep))))]},
+	$(if $(PROJECT_MOD),{mod$(comma) {'$(PROJECT_MOD)'$(comma) []}}$(comma),)
+	{env, $(subst \,\\,$(PROJECT_ENV))}$(if $(findstring {,$(PROJECT_APP_EXTRA_KEYS)),$(comma)$(newline)$(tab)$(subst \,\\,$(PROJECT_APP_EXTRA_KEYS)),)
+]}.
+endef
```
essen, line 53: Not sure why this was redeclared either.

core/elixir.mk
```diff
@@ -0,0 +1,334 @@
...
+			{ok, PackageInfo} =
+				case hex_repo:get_package(hex_core:default_config(), atom_to_binary(Name)) of
+					{ok, {200, _RespHeaders, Decoded}} ->
+						LFirst(Decoded, fun(#{version := Vsn}) -> 'Elixir.Version':'match?'(Vsn, Req) end);
+					Other ->
+						io:format(standard_error, "Unexpected response for Dep ~p", [Name]),
+						erlang:halt(1)
+				end,
+			maps:get(version, PackageInfo)
+		end,
```
essen, line 118: This should probably be moved to the hex part so we can use it for Erlang deps as well.

essen:
Is the thing that transforms paths like Elixir/MyApp/MyMod.ex into module name `Elixir.MyApp.MyMod` something the compiler does, or am I just blind and I missed it?

artman41:
@essen had some time recently so will be picking this one back up 👍

test/log/console.log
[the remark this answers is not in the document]
artman41: removed

core/elixir.mk
```diff
@@ -0,0 +1,334 @@
...
+endef
+
+ebin/$(PROJECT).app:: $(EX_FILES)
+	$(gen_verbose) $(if $(strip $(EX_FILES)),$(call compile_ex,$(EX_FILES)))
+# Older git versions do not have the --first-parent flag. Do without in that case.
+	$(eval GITDESCRIBE := $(shell git describe --dirty --abbrev=7 --tags --always --first-parent 2>/dev/null \
+		|| git describe --dirty --abbrev=7 --tags --always 2>/dev/null || true))
+	$(eval MODULES := $(patsubst %,'%',$(sort $(notdir $(basename \
+		$(filter-out $(ELIXIRC_EXCLUDE_PATHS), $(ERL_FILES) $(CORE_FILES) $(BEAM_FILES)))))))
+	$(eval MODULES := $(call UNIQUE,$(MODULES) $(foreach file, \
```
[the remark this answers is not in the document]
artman41: I think that's a fair point, likely more safety/fear on my part 👍

core/deps.mk
artman41: likely a misunderstanding on my part - I'd assumed they needed adding to the `LOCAL_DEPS_DIRS` var since I was adding them to `LOCAL_DEPS`

core/deps.mk
```diff
@@ -778,14 +780,16 @@ define dep_fetch_ln
 	ln -s $(call dep_repo,$(1)) $(DEPS_DIR)/$(call dep_name,$(1));
 endef
 
+HEX_REPO = https://repo.hex.pm
```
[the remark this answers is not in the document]
artman41: My only pushback against that would be for if someone is using an internal mirror within a corporate setting - it's likely they'd want to instead have the hex side of `erlang.mk` pointing at whatever their mirror is

looking back at it, I think I'd likely want to change this to be

```make
HEX_REPO ?= https://repo.hex.pm
```

core/erlc.mk
```diff
@@ -86,7 +86,7 @@ define app_file
 endef
 endif
 
-app-build: ebin/$(PROJECT).app
+app-build:: ebin/$(PROJECT).app
```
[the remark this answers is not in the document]
artman41: it's a fair point, I've also removed the `$(PROJECT).d` target & the body of the `app::` target above this

core/elixir.mk
artman41: The reason for this redeclaration was to have .app files generate where the `mod` may actually be an elixir module

Currently, the `mod` prop only gets inserted if an erlang file exists.

In efforts to separate `erlc.mk` from `elixir.mk`, I opted to redefine the macro - if you'd prefer, I could add a condition to the definition in `erlc.mk`?

core/elixir.mk
artman41: Assuming you meant the `GetDeps` fun, I've moved that to `elixir_get_deps.erl` within elixir.mk - if you meant the macro itself, let me know and I'll move it over 👍

artman41:
> Is the thing that transforms paths like Elixir/MyApp/MyMod.ex into module name `Elixir.MyApp.MyMod` something the compiler does, or am I just blind and I missed it?

Compiler magic 🪄

essen:
I will try to make this work with RabbitMQ CLI which is (currently) in Elixir. Could you rebase to run CI now that I fixed it? Also you will need to edit the workflow to add/change the new suites here https://github.com/ninenines/erlang.mk/blob/master/.github/workflows/ci.yaml#L27

essen:
I'm looking to merge this now. No need to rebase, I am probably going to cherry-pick parts of the code while I try to make things work.

essen:
OK first step is obviously to merge the version resolver, this will also be good to have for non-Elixir projects.

Sorry I'm taking so long, I tried hard to refactor autopatch into something more pluggable (one step detect one step patch based on detection) but I'll have to leave it for later.

core/deps.mk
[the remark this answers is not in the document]
essen: The proxy can be configured via hex_core.

core/elixir.mk
```diff
@@ -0,0 +1,269 @@
...
+	$(verbose) $(MAKE) -C $(DEPS_DIR)/elixir -f Makefile.orig compile Q="$(verbose)"
+endif
+
+# We need the original makefile so that we can compile the elixir compiler
+autopatch-elixir::
+	$(verbose) cp $(DEPS_DIR)/elixir/Makefile $(DEPS_DIR)/elixir/Makefile.orig
+	$(verbose) sed 's|"$$(MAKE)"|"$$(MAKE)" -f $$(CURDIR)/Makefile.orig|g' -i $(DEPS_DIR)/elixir/Makefile.orig
+
+ifneq ($(USES_ELIXIR),)
+elixir-check-crypto:
```
essen, line 252: I don't think we need this check, at least not anymore. Since we use hex_core for fetching, and hex_core fetches over TLS, it necessarily has crypto available.

Even if we didn't I don't think we want to encourage the practice of splitting Erlang/OTP into multiple "optional" packages like some distros do. You are expected to have the full Erlang/OTP when you use Erlang.mk.

core/elixir.mk
```diff
@@ -0,0 +1,269 @@
...
+
+define compile_ex
+$(elixirc_verbose) ERL_COMPILER_OPTIONS="[$(call comma_list,$(patsubst '%',%,$(patsubst +%,%,$(filter +%,$(ELIXIRC_OPTS))))), {pa, \"ebin/\"}, {i, \"include/\"}]" $(ELIXIRC) \
+	--verbose $(if $(IS_DEP),,$(if $(filter -Werror,$(ELIXIRC_OPTS)),--warnings-as-errors)) \
+	-o ebin/ \
+	$(filter-out $(ELIXIRC_EXCLUDE_PATHS),$(ELIXIR_COMPILE_FIRST_PATHS)) $(1) 
+endef
+
+# Currently doesn't account for nested modules
+define get_elixir_mod
```
essen, line 200: I think we should just trust the Elixir compiler to do the right thing, and wildcard the modules it compiled in `ebin/`. They all start with `Elixir.` so it should be easy enough.

core/elixir.mk
essen: I am changing this to resolving PROJECT_MOD first, before defining the app_file. This allows to merge all 3 separate app_file definitions into one.

core/elixir.mk
essen: I changed it so that we call the Elixir compiler from Erlang code and get the list of compiled modules out of that.

essen:
I pretty much have to merge some of the Elixir stuff into erlc.mk (and some stuff in deps.mk). Otherwise it just ends up conflicting during execution leading to things not getting rebuilt and whatnot.

core/elixir.mk
```diff
@@ -0,0 +1,269 @@
...
+fi
+endef
+
+define SHELL_ASSERT
+$(if $(verbose),,$(info [SHELL_ASSERT] $(call SHELL_ASSERT_,$(1),$(2)))) $(call SHELL_ASSERT_,$(1),$(2))
+endef
+
+define dep_autopatch_mix
+	$(MAKE) $(ELIXIRC) hex-core || exit 1; \
+	sed 's|\(defmodule.*do\)|\1\ntry do\nCode.compiler_options(on_undefined_variable: :warn)\nrescue _ -> :ok\nend|g' -i $(DEPS_DIR)/$(1)/mix.exs; \
```
essen, line 173: Do you know why this line is necessary? Mix magic?

core/elixir.mk
artman41: My only suggestion to this would be to mention crypto as an explicit dependency / add a description of having a full Erlang/OTP installation to the user guide and github README 👍

core/elixir.mk
essen: Thanks, I get that. But I meant is using an undefined variable in Elixir a normal OK thing to do? I don't know Elixir so just wondering.

core/elixir.mk
essen: We can do that yes.

core/elixir.mk
artman41: Honestly, I'm unsure 

I only found the issue when testing with a dep - I'm not sure whether it was phoenix or another elixir library

I'll try and have a look in a bit to see if I can find the library again 👍

essen: It did break tests when I removed it.

essen:
The PR at https://github.com/ninenines/erlang.mk/pull/1020 is almost ready, just a few details and more tests to add. It completely replaces Mix in RabbitMQ in https://github.com/rabbitmq/rabbitmq-server/pull/12922 except for running tests.

essen:
Closing in favor of the PR mentioned above.

Thank you very much for the hard work on this!!

PR #982, opened 2023-06-20
#981 - Fix Verbosity Regression

### LINKED ISSUE #981: autopatching is overly verbose (missing $(verbose))
Somewhere in the spree of commits you merged @essen, a change to remove autopatching for rabbitmq also removed the `$(verbose)` preceding the bash

Before
https://github.com/ninenines/erlang.mk/blob/a1d7135f1349735569016adcb3e185be58c44042/core/deps.mk#L768-L788

After
https://github.com/ninenines/erlang.mk/blob/e1e3d15a1b211fc60b83557a2ea54ea3741f3e80/core/deps.mk#L841-L847

This results in a lot of debug info being printed to the terminal during the `deps` step as all the autopatching is executed

artman41:
Fixes #981

### DIFF
--- a/core/deps.mk
+++ b/core/deps.mk
@@ -839,7 +839,7 @@ endif
 .PHONY: autopatch-$(call dep_name,$1)
 
 autopatch-$(call dep_name,$1)::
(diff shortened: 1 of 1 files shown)


essen:
Merged, thanks!

Issue #970, opened 2023-03-17
Feature: Additional template modifiers


artman41:
Currently, we're limited to templates that will create files under either `src/` or `$app/src/`

I think it'd be useful if templates could have a defined relative path and file extension so that we could create hrl templates & test templates.

e.g

**Template**
```
tpl_hrl_dir = include/
tpl_hrl_ext = .hrl
define tpl_hrl
-ifndef(HRL_$(n)).
-define(HRL_$(n), 1).

-endif.
endef
```

**Execution**
```
make new t=hrl n=my_file
```

`include/my_file.hrl`
```
-ifndef(HRL_my_file).
-define(HRL_my_file, 1).

-endif.
```

essen:
That's a good idea. Not sure about how it should be implemented but patch is welcome!

essen:
Closing in favor of the PR. Templates have been reworked, now all that's needed is updating the PR to include the few extra templates and template changes.

PR #983, opened 2023-07-04
#970 - Allow for additional template Modifiers

artman41:
Implements #970

* Templates now exist under templates/
* Templates can now have a directory and extension defined
* Added hrl template
* Updated supervisor template to include Child Specs
* Made some bootstrap templates generateable via new

### DIFF
--- a/build.config
+++ b/build.config
@@ -19,6 +19,14 @@ core/docs
 core/rel
 core/test
 core/compat
+core/templates
+
+# Templates
+templates/bootstrap
+templates/release
+templates/erlang
+templates/cowboy
+templates/ranch
 
 # Plugins.
 plugins/asciidoc
--- a/core/templates.mk
+++ b/core/templates.mk
@@ -1,0 +1,60 @@
+new-app:
+ifndef in
+	$(error Usage: $(MAKE) new-app in=APP)
+endif
+ifneq ($(wildcard $(APPS_DIR)/$in),)
+	$(error Error: Application $in already exists)
+endif
+	$(eval p := $(in))
+	$(if $(shell echo $p | LC_ALL=C grep -x "[a-z0-9_]*"),,\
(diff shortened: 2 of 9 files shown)


essen:
Please also rebase this one.

artman41:
@essen rebased this one 👍

essen:
Thank you!

essen:
I think I would take a different approach. It would be good if template files were completely outside of Makefiles in the Erlang.mk source, as well as in third party plugins. This would make editing and testing them simpler. Of course the compiled Erlang.mk file should include them inside the file, not separately. So it all comes down to being able to declare templates that will be loaded from a file if the corresponding variable doesn't exist, and making the Erlang.mk build embed the files inside variables.

If templates become core functionality, then the entire bootstrap plugin probably should be core as well. Not much remains.

I will work on something.

essen:
I have opened https://github.com/ninenines/erlang.mk/pull/1013 for my version of the thing. Almost all templates including internal have been replaced. Makefiles remain. Some plain substitution happens for module names and whatnot. Erlang.mk still embeds templates but it's possible to use `tpl_my_template $(file < templates.my_template)` to define a template, I will probably port the Cowboy templates as a third party plugin using this method (so these will never be embedded). There's still some more work to do but it's getting there. Once the core of the changes is done I'll leave it up to you to update the templates, add the hrl one and any other small things.

By the way do you remember why you moved them to core?

essen:
I have merged my changes: https://github.com/ninenines/erlang.mk/commit/a6f9a450af99a7a95848dde671de658bb53a43dc

When you have time, please add the additional templates and template fixes you included in this PR or were thinking of including.

essen:
Closing since it's mostly done. If you want you can open a new PR adding new templates. Thanks!

PR #986, opened 2023-07-07
Update erlang downloads links on docs

yskelg:
Old Erlang Download links are broken. So, Update new download links on docs.


essen:
Merged, thanks!

PR #1005, opened 2024-09-25
Fix conversion of git_subdir to git-subfolder

### LINKED ISSUE #1004: case_clause in dep_autopatch_rebar.erl when transforming git_subdir
Rebar 3.14 added `git_subdir` to its supported dependency formats. See https://rebar3.org/docs/configuration/dependencies/. These have 4 elements. This causes `dep_autopatch_rebar.erl` to fail with the following error:

```
Runtime terminating during boot ({{case_clause,{git_subdir,"https://github.com/whatsapp/eqwalizer.git",{branch,"main"},"eqwalizer_support"}},...
```

rebar.config references eqWAlizer, following the instructions here: https://github.com/whatsapp/eqwalizer?tab=readme-ov-file#using-it-with-rebar3-projects

In the case of eqwalizer, because it's only used at build-time, I can probably work around this with a rebar profile that erlang.mk will ignore. But with other `git_subdir` dependencies, such as opentelemetry-erlang, this won't be feasible.

rlipscombe:
Fixes #1004

### DIFF
--- a/core/deps.mk
+++ b/core/deps.mk
@@ -444,6 +444,12 @@ define dep_autopatch_rebar.erl
 			GetHexVsn3Common(N, NP, S0);
 		(N, NP, S) -> {N, {hex, NP, S}}
 	end,
+	ConvertCommit = fun
+		({branch, C}) -> C;
+		({ref, C}) -> C;
+		({tag, C}) -> C;
+		(C) -> C
+	end,
 	fun() ->
 		File = case lists:keyfind(deps, 1, Conf) of
 			false -> [];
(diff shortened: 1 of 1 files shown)


essen:
Do you have an open source project that uses git_subdir that I can use to confirm that this works?

rlipscombe:
https://github.com/rlipscombe/subdir_example and https://github.com/rlipscombe/subdir_example_rebar should be enough to reproduce the problem. I cobbled them together in a hurry, though, so let me know if they need any extra work.

essen:
Thank you, will check them out this week.

PR #1013, opened 2024-11-21
Improve templates

essen:
[Move templates outside the source .mk files](https://github.com/ninenines/erlang.mk/pull/1013/commits/e6e3f82bb08c78303335923f5d7f32d6fde73c50)
Templates now no longer use Make variables for substitution
but instead replace strings with their equivalent:

```
template_name: Corresponds to n=template_name
project_name:  Corresponds to $(PROJECT) or in=project_name
```

This allows defining templates outside of Makefiles. For
example an external plugin could define their templates
in templates/my_template.erl and then have the following
in the included Makefile:

```
tpl_my_template = $(file < $(DEPS_DIR)/my_plugin/templates/my_template.erl)
```

By default the created file will be in src/template_name.erl.
This can be overriden with the tplp_* variable:

```
tplp_my_template = src/model/my_template.erl
```

Substitution is applied both to the template contents and
to its path.

My version of #983


essen:
Merged!!

PR #1018, opened 2024-11-26
Make $(ERL) faster

essen:
Using "-mode minimal" and restricting processes and sockets to a maximum of 1024. Minimal mode has been available for many years. The downside is that some things require us to start inet_db manually, such as Hex.


essen:
Merged!

PR #1020, opened 2024-12-05
Native Elixir support

essen:
Heavily modified version of https://github.com/ninenines/erlang.mk/pull/979

### DIFF
--- a/core/deps.mk
+++ b/core/deps.mk
@@ -118,6 +118,8 @@ dep_name = $(call query_name,$(1))
 # Application directories.
 
 LOCAL_DEPS_DIRS = $(foreach a,$(LOCAL_DEPS),$(if $(wildcard $(APPS_DIR)/$a),$(APPS_DIR)/$a))
+# Elixir is handled specially as it must be built before all other deps
+# when Mix autopatching is necessary.
 ALL_DEPS_DIRS = $(addprefix $(DEPS_DIR)/,$(foreach dep,$(filter-out $(IGNORE_DEPS),$(BUILD_DEPS) $(DEPS)),$(call query_name,$(dep))))
 
 # When we are calling an app directly we don't want to include it here
@@ -210,9 +212,11 @@ endif
 ifneq ($(SKIP_DEPS),)
 deps::
 else
-deps:: $(ALL_DEPS_DIRS) apps clean-tmp-deps.log | $(ERLANG_MK_TMP)
-ifneq ($(ALL_DEPS_DIRS),)
-	$(verbose) set -e; for dep in $(ALL_DEPS_DIRS); do \
+ALL_DEPS_DIRS_TO_BUILD = $(if $(filter-out $(DEPS_DIR)/elixir,$(ALL_DEPS_DIRS)),$(filter-out $(DEPS_DIR)/elixir,$(ALL_DEPS_DIRS)),$(ALL_DEPS_DIRS))
+
+deps:: $(ALL_DEPS_DIRS_TO_BUILD) apps clean-tmp-deps.log | $(ERLANG_MK_TMP)
+ifneq ($(ALL_DEPS_DIRS_TO_BUILD),)
+	$(verbose) set -e; for dep in $(ALL_DEPS_DIRS_TO_BUILD); do \
 		if grep -qs ^$$dep$$ $(ERLANG_MK_TMP)/deps.log; then \
 			:; \
 		else \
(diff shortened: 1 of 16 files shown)


essen:
We're green! Just have to do autopatch cleanup, move things around and review.

essen:
Something that this doesn't currently do is skip rebuilding.

essen:
Once all tests pass again in RabbitMQ this can be merged. But perhaps leave it hanging until after Christmas.

essen:
This broke Ranch use within tests somehow.

core/deps.mk
```diff
@@ -854,30 +901,39 @@ $(DEPS_DIR)/$(call query_name,$1): | $(if $(filter hex,$(call query_fetch_method
...
 ifeq ($(filter $1,$(NO_AUTOPATCH)),)
-	$(verbose) $$(MAKE) --no-print-directory autopatch-$(DEP_NAME)
+	$(verbose) $$(MAKE) --no-print-directory autopatch-$(DEP_NAME) AUTOPATCH_METHOD=`$(call dep_autopatch_detect,$1)`
 endif
 
 .PHONY: autopatch-$(call query_name,$1)
 
+ifeq ($1,elixir)
+autopatch-elixir::
+	ln -s lib/elixir/ebin $(DEPS_DIR)/elixir/
```
essen, line 911: $(verbose)

essen:
For eunit we need to filter out the Elixir.*.beam files.

core/elixir.mk
```diff
@@ -0,0 +1,145 @@
+# Copyright (c) 2024, Tyler Hughes <artman41@gmail.com>
```
artman41, line 1: sorry to be awkward Essen, any chance I could get this swapped to my dev email `tyler@tylerhughes.dev`?

No worries if not, I appreciate it's a tiny thing 😅

essen: Will do. I think I used the email from your PR's commits. If I forget a PR updating it is of course welcome.

artman41:
I've been playing around with the branch tonight essen, found some inconsistencies when using Elixir dependencies so created a 'fixes' branch - PR is at #1021

One of the commits adds support for Elixir 1.13 which supports OTP 23 - I'm happy to remove this commit though if you'd rather not have the backwards compat

essen:
@artman41 The branch looks good as far as RabbitMQ is concerned. Please do a final check before I squash and merge. Thank you!!

core/deps.mk
lhoguin: This still needs to be fixed.

essen:
There's a couple comments left in this PR that are worth looking at again before merging.

core/elixir.mk
```diff
@@ -0,0 +1,161 @@
+# Copyright (c) 2024, Tyler Hughes <tyler@tylerhughes.dev>
+# Copyright (c) 2024, Loïc Hoguin <essen@ninenines.eu>
+# This file is part of erlang.mk and subject to the terms of the ISC License.
+
+ELIXIR ?= $(if $(filter elixir,$(BUILD_DEPS) $(DEPS)),dep,system)
+export ELIXIR
+
+ifeq ($(ELIXIR),system)
+# We expect 'elixir' to be on the path.
+# @todo Only if there are EX_FILES
```
essen, line 10: This should probably be resolved before merging as well.

essen:
I think we also want to make sure that setting e.g. `ELIXIR = disable` completely disables Elixir support.

core/elixir.mk
essen: This can only be resolved by converting the Hex version resolver to use an Erlang equivalent of Elixir's Version.match? so I guess I am doing that now...

essen: It appears that https://github.com/jelly-beam/verl works just fine. It supports `>= 1.8.0 and < 3.0.0` as well, which e.g. Cowboy uses.

So it is just a matter of fetching `verl`. I guess we could have a meta project whose only purpose is to depend on both `hex_core` and `verl`; or fetch them individually depending on needs. This might make the most sense, we only need `verl` for autopatch, whereas `hex_core` is used for other Hex related functions.

Alternatively perhaps @wojtekmach would like to have `verl` a part of `hex_core` (either as a dep or vendored in) so that we have the full functionality out of the box. Just wondering.

artman41:
> @artman41 The branch looks good as far as RabbitMQ is concerned. Please do a final check before I squash and merge. Thank you!!

Looks good so far I think, nothing obvious I'm seeing in terms of issues 👍

core/elixir.mk
essen: OK I think we can restrict the Elixir resolver to Elixir, and have an Erlang resolver for Erlang later (perhaps use what's in Rebar3 directly). So this is not blocking.

I will move the Hex resolver to elixir.mk.

essen:
I will add a test to check that we can disable Elixir entirely.

essen:
OK so I don't want a `readlink` command run every time Make is run, as that slows down everything when not needed. If there's EX_FILES we can enable Elixir. If Elixir is in deps we can use that one. But if neither is true, AND we want to use Elixir deps, then I think we must require `ELIXIR = system` to be set. That's probably rare, and OK.

core/elixir.mk
wojtekmach: @essen agreed, I created a ticket to track it: https://github.com/hexpm/hex_core/issues/157

essen:
Most recent commit should be it, after that a squash and merge is all that's left. Plus some documentation I guess (at least explaining how to disable Elixir, or enable it for deps), and updating the changelog. I plan to get it merged Monday/Tuesday.

essen:
OK there's still a problem with Elixir activation. Elixir project in RabbitMQ fails saying to enable Elixir.

essen:
If Elixir is disabled we shouldn't try to autopatch Mix projects (some projects have both Erlang and Elixir so that's OK, such as `jose`).

essen:
OK there's now a fallback to Rebar3 when Elixir is disabled, so `jose` can be used without Elixir enabled again.

essen:
@artman41 Final commit has the beginning of a docs page.

I will merge this tomorrow if everything is OK.

essen:
Merged!! Thanks!

PR #1021, opened 2025-01-09
Native Elixir Fixes

artman41:
* **Cause failed compilation of .ex files to prevent creation of .app file**
    * **Replace wildcard for ex beams for a shell command** 
    * _To prevent failed `.ex` compilation from continuing to create a `$(PROJECT).app` file, I've moved the elixir generation code to be ran by the actual shell. Doing this though meant the Module names of the compiled `.ex` files needed to be found separately. To solve this, I've tried to use a shell command to manually build the Module name based on the `__aliases__` attribute in the elixir ast of the code_
* **Allow support for Elixir v1.13 (supported by OTP 23)**
     * _I'm happy to remove this commit from the branch if you don't want to support OTP 23 essen_
* **Replace my email**
* **Account for .ex files which reference mix.exs config**
    * _I noticed when using the dependency `mint` that it referenced the `Mix.Project.config()` at compile time, hence the need to load the `mix.exs` project during `.ex` compilation_
* **Simplify Ex compilation failure**
    * _As suggested, modified the `compile_ex.erl` to return the compiled modules for easier var binding. The erl code will now redirect the errors printed by `'Elixir.Kernel.ParallelCompiler':compile_to_path/2` to `stderr` & print to `stdout` only for the module names or the word `_ERROR_` on a failed compilation_

### DIFF
--- a/core/elixir.mk
+++ b/core/elixir.mk
@@ -35,7 +35,7 @@ define dep_autopatch_mix.erl
 	Application = try Mod:application() catch error:undef -> [] end,
 	StartMod = case lists:keyfind(mod, 1, Application) of
 		{mod, {StartMod0, _StartArgs}} ->
-			StartMod0;
+			atom_to_list(StartMod0);
 		_ ->
 			""
 	end,
--- a/core/elixir.mk
+++ b/core/elixir.mk
@@ -25,7 +25,11 @@ define dep_autopatch_mix.erl
 	{ok, _} = application:ensure_all_started(elixir),
 	{ok, _} = application:ensure_all_started(mix),
 	MixFile = <<"$(call core_native_path,$(DEPS_DIR)/$1/mix.exs)">>,
-	[{Mod, Bin}] = elixir_compiler:file(MixFile, fun(_File, _LexerPid) -> ok end),
+	{Mod, Bin} =
+		case elixir_compiler:file(MixFile, fun(_File, _LexerPid) -> ok end) of
+			[{T = {_, _}, _CheckerPid}] -> T
+			[T = {_, _}] -> T;
+		end,
 	{module, Mod} = code:load_binary(Mod, binary_to_list(MixFile), Bin),
 	Project = Mod:project(),
 	Application = try Mod:application() catch error:undef -> [] end,
--- a/core/elixir.mk
+++ b/core/elixir.mk
@@ -143,12 +143,15 @@ define compile_ex.erl
 	ModCode = list_to_atom("Elixir.Code"),
 	ModCode:put_compiler_option(ignore_module_conflict, true),
 	ModComp = list_to_atom("Elixir.Kernel.ParallelCompiler"),
-	case ModComp:compile_to_path([$(call comma_list,$(patsubst %,<<"%">>,$(EX_FILES)))], <<"ebin/">>) of
-		{ok, Modules, _} ->
-			halt(0);
-		{error, [], _WarnedModules} ->
-			halt(0);
-		{error, _ErroredModules, _WarnedModules} ->
-			halt(1)
-	end
+	ModMixProject = list_to_atom("Elixir.Mix.Project"),
+	ModMixProject:in_project($(PROJECT), ".", [], fun(_MixFile) ->
+		case ModComp:compile_to_path([$(call comma_list,$(patsubst %,<<"%">>,$(EX_FILES)))], <<"ebin/">>) of
+			{ok, Modules, _} ->
+				halt(0);
+			{error, [], _WarnedModules} ->
+				halt(0);
+			{error, _ErroredModules, _WarnedModules} ->
+				halt(1)
+		end
+	end)
 endef
--- a/core/elixir.mk
+++ b/core/elixir.mk
@@ -138,41 +138,21 @@ define dep_autopatch_mix
 endef
 
 define compile_ex.erl
-	{ok, _} = application:ensure_all_started(elixir),
+	{ok, _} = application:ensure_all_started(elixir),	
 	{ok, _} = application:ensure_all_started(mix),
 	ModCode = list_to_atom("Elixir.Code"),
 	ModCode:put_compiler_option(ignore_module_conflict, true),
 	ModComp = list_to_atom("Elixir.Kernel.ParallelCompiler"),
 	ModMixProject = list_to_atom("Elixir.Mix.Project"),
+	erlang:group_leader(whereis(standard_error), self()),
 	ModMixProject:in_project($(PROJECT), ".", [], fun(_MixFile) ->
 		case ModComp:compile_to_path([$(call comma_list,$(patsubst %,<<"%">>,$(EX_FILES)))], <<"ebin/">>) of
 			{ok, Modules, _} ->
-				halt(0);
-			{error, [], _WarnedModules} ->
+				lists:foreach(fun(E) -> io:format(user, "~p ", [E]) end, Modules),
 				halt(0);
 			{error, _ErroredModules, _WarnedModules} ->
+				io:format(user, "_ERROR_", []),
 				halt(1)
 		end
 	end)
-endef
-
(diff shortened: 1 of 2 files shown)


core/erlc.mk
```diff
@@ -306,11 +306,11 @@ endef
 ebin/$(PROJECT).app:: $(ERL_FILES) $(CORE_FILES) $(wildcard src/$(PROJECT).app.src) $(EX_FILES)
 	$(eval FILES_TO_COMPILE := $(filter-out $(EX_FILES) src/$(PROJECT).app.src,$?))
 	$(if $(strip $(FILES_TO_COMPILE)),$(call compile_erl,$(FILES_TO_COMPILE)))
-	$(if $(filter $?,$(EX_FILES)),$(elixirc_verbose) $(eval MODULES := $(shell $(call erlang,$(call compile_ex.erl,$(EX_FILES))))))
+	$(if $(filter $?,$(EX_FILES)),$(elixirc_verbose) $(call erlang,$(call compile_ex.erl,$(EX_FILES))))
 # Older git versions do not have the --first-parent flag. Do without in that case.
 	$(eval GITDESCRIBE := $(shell git describe --dirty --abbrev=7 --tags --always --first-parent 2>/dev/null \
 		|| git describe --dirty --abbrev=7 --tags --always 2>/dev/null || true))
-	$(eval MODULES := $(MODULES) $(patsubst %,'%',$(sort $(notdir $(basename \
+	$(eval MODULES := $(MODULES) $(notdir $(basename $(wildcard ebin/Elixir.*))) $(patsubst %,'%',$(sort $(notdir $(basename \
```
essen, line 313: You can't do that, the wildcard function operates on cache that is updated at the start of executing the target, so it won't know about the new Elixir beam files.

essen:
I need more info about the two most recent commits because I don't know what they do.

core/erlc.mk
artman41: good point, leave it with me and I'll double check against what I have locally - could be the case that I need to rollback that change

artman41:
@essen tried a different way of getting the elixir beams, let me know what you think

I'll add an explanation under the commits in the PR description to say why it's needed 👍

essen:
I don't think you need to go that far. It is much better to get the modules list from the compilation, we only need a way to signal success or failure. That could be signaled as a word in the output, which we would then check, before making `MODULES` be everything except that (first or last) word. The check would error out if failure is indicated.

See word related functions in https://www.gnu.org/software/make/manual/html_node/Text-Functions.html especially firstword, wordlist.

artman41:
> I don't think you need to go that far. It is much better to get the modules list from the compilation, we only need a way to signal success or failure. That could be signaled as a word in the output, which we would then check, before making `MODULES` be everything except that (first or last) word. The check would error out if failure is indicated.
> 
> See word related functions in https://www.gnu.org/software/make/manual/html_node/Text-Functions.html especially firstword, wordlist.

I feel dumb - I forgot entirely about `firstword`, glad you're able to offer a different perspective 😄 

I'll revert that and get a change pushed up now 👍

core/elixir.mk
```diff
@@ -134,12 +138,21 @@ define dep_autopatch_mix
 endef
 
 define compile_ex.erl
-	{ok, _} = application:ensure_all_started(elixir),
+	{ok, _} = application:ensure_all_started(elixir),
```
essen, line 141: Bad whitespace.

core/elixir.mk
```diff
@@ -134,12 +138,21 @@ define dep_autopatch_mix
...
+	{ok, _} = application:ensure_all_started(elixir),	
 	{ok, _} = application:ensure_all_started(mix),
 	ModCode = list_to_atom("Elixir.Code"),
 	ModCode:put_compiler_option(ignore_module_conflict, true),
 	ModComp = list_to_atom("Elixir.Kernel.ParallelCompiler"),
-	{ok, Modules, _} = ModComp:compile_to_path([$(call comma_list,$(patsubst %,<<"%">>,$(EX_FILES)))], <<"ebin/">>),
-	lists:foreach(fun(E) -> io:format("~p ", [E]) end, Modules),
-	halt()
+	ModMixProject = list_to_atom("Elixir.Mix.Project"),
+	erlang:group_leader(whereis(standard_error), self()),
```
essen, line 147: Could you explain this line?

core/erlc.mk
```diff
@@ -306,13 +306,15 @@ endef
...
-	$(if $(filter $?,$(EX_FILES)),$(elixirc_verbose) $(eval MODULES := $(shell $(call erlang,$(call compile_ex.erl,$(EX_FILES))))))
+	$(if $(filter $?,$(EX_FILES)),$(elixirc_verbose) $(eval MODULES := $(shell $(call erlang,$(call compile_ex.erl)))))
+	$(eval ELIXIR_COMP_FAILED := $(if $(filter _ERROR_,$(firstword $(MODULES))),true,false))
 # Older git versions do not have the --first-parent flag. Do without in that case.
 	$(eval GITDESCRIBE := $(shell git describe --dirty --abbrev=7 --tags --always --first-parent 2>/dev/null \
 		|| git describe --dirty --abbrev=7 --tags --always 2>/dev/null || true))
 	$(eval MODULES := $(MODULES) $(patsubst %,'%',$(sort $(notdir $(basename \
 		$(filter-out $(ERLC_EXCLUDE_PATHS),$(ERL_FILES) $(CORE_FILES) $(BEAM_FILES)))))))
 ifeq ($(wildcard src/$(PROJECT).app.src),)
+	$(verbose) if $(ELIXIR_COMP_FAILED); then exit 1; fi;
```
essen, line 317: I do not understand why the failure check is done here instead of immediately after trying to compile files. Also why it is checked only when there's no .app.src file.

core/erlc.mk
```diff
@@ -306,13 +306,15 @@ endef
 ebin/$(PROJECT).app:: $(ERL_FILES) $(CORE_FILES) $(wildcard src/$(PROJECT).app.src) $(EX_FILES)
 	$(eval FILES_TO_COMPILE := $(filter-out $(EX_FILES) src/$(PROJECT).app.src,$?))
 	$(if $(strip $(FILES_TO_COMPILE)),$(call compile_erl,$(FILES_TO_COMPILE)))
-	$(if $(filter $?,$(EX_FILES)),$(elixirc_verbose) $(eval MODULES := $(shell $(call erlang,$(call compile_ex.erl,$(EX_FILES))))))
+	$(if $(filter $?,$(EX_FILES)),$(elixirc_verbose) $(eval MODULES := $(shell $(call erlang,$(call compile_ex.erl)))))
```
essen, line 309: Rather than removing the argument here I would rather `compile_ex.erl` use `$1` instead of `$(EX_FILES)`.

essen (changes requested):
A few questions. Don't worry too much about doing the changes, if the questions are solved I can do the small tweaks myself.

core/elixir.mk
artman41: Elixir calls `io:format` internally as part of the `Mod:in_project/4` function which means the output gets captured by `$(shell ...)`

I managed to avoid the capture from happening by redirecting `io:format` to `standard_error` via the group_leader

core/erlc.mk
artman41: happy to move it - the failure was there mainly because if a module fails to compile, it'll prevent the `.app` from being generated correctly because the error text will be in the `modules` section of the app resource

*technically* if there was an `app.src` file which defined the modules manually, it should succeed

core/erlc.mk
artman41: sounds good 👍

core/erlc.mk
essen: Yeah but we want neither to succeed. We don't want to generate the .app, regardless of method, and just fail right after the compilation of Elixir files failed. Like for erlc.

PR #1022, opened 2025-01-10
Fix execution order of 'make app shell'

artman41:


essen:
Currently `make shell` doesn't depend on `app` at all, so this changes the behavior. I don't know if we want this. This should only be an issue if you have parallel Make (`-j`) enabled. This is easily solved via `make app && make shell`.

artman41:
I'll close it for now in that case 😄 

I put it in because I was having some issues when trying to run `make app shell` using otp 27 😅

PR #1033, opened 2025-04-27
Refine postprocessing of project name for use in man pages

frou:
Hi there, it's really nice that generation of man pages is possible (`make install-docs`).

When viewing them for Cowboy, I noticed that the project name is messed up:

Screenshot 2025-04-27 at 12 35 37

That seems to be because a non-portable `sed` feature is used to capitalise the first letter. I've replaced that with a call to `awk` that [should be portable.](https://pubs.opengroup.org/onlinepubs/9699919799/utilities/awk.html)


essen:

PR #1035, opened 2025-06-30
Collection of test fixes

dumbbell:
Here is a summary of the fixes:
* test: Set the branch name in `git init`
* test: Bump some dependencies that fail to build with Erlang/OTP 28
* test/core_app.mk: Redirect stderr to stdout when tests grep for parse_transform output
* test/core_app.mk: Use https://github.com/binarytemple/hypernumbers
* test/core_query.mk: `gun` master branch depends on cowlib 2.15.0
* test/plugin_hex.mk: Use the name "nc" for the netcat command

With them, I can run the whole testsuite successfully on FreeBSD with Erlang/OTP 28.

PR #1036, opened 2025-07-08
Optimize `dep_fetch_hex` on Windows

dumbbell:
> WIP: the pull request is based on another topic branch, only the last commit is relevant in the branch will be rebased once the other commits are reviewed and merged.

We do this by avoiding calling `core_native_path` in dep_fetch_hex many times.

On Windows, this executes a shell to determine the native path, which is very expensive. The result is always the same, so make sure `core_native_path` is called once instead.

For example, this saves a significant amount of time when compiling RabbitMQ on Windows.

### DIFF
--- a/core/core.mk
+++ b/core/core.mk
@@ -59,11 +59,6 @@ gen_verbose_esc_0 = @echo " GEN   " $$@;
 gen_verbose_esc_2 = set -x;
 gen_verbose_esc = $(gen_verbose_esc_$(V))
 
-# Temporary files directory.
-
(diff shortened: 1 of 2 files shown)


dumbbell:
The patch is ready. I don’t mark it as ready for review yet because it is based on the branch for #1035.

essen:
Closing in favor of https://github.com/ninenines/erlang.mk/pull/1059

This PR is already merged there with a few tweaks and will be merged into main when the next set of changes is ready.

dumbbell:
Thank you!

PR #1042, opened 2025-08-29
Fix autopatch of erlang-rocksdb

essen:
Thanks Ilya Khaprov for the fix.


essen:
Build will probably fail because of https://github.com/facebook/rocksdb/issues/13365

PR #1046, opened 2026-05-20
Fix errors in CI

Zerpet:
The test fetches `gun` master branch, which now requires Cowlib 2.16.1. The test was expecting Cowlib 2.15.0.

Adding `-f` to `lockf` to preserve the lock file after `lockf` exits, preserving the concurrency guard rail.
I'm unsure if this flag is relevant to CI failures, but I couldn't get tests to pass locally without it.

Using 8-character passwords in tests for hexpm DB users, because that's the minimum required by the server.

### DIFF
--- a/core/deps.mk
+++ b/core/deps.mk
@@ -318,7 +318,7 @@ define maybe_flock
 	if command -v flock >/dev/null; then \
 		flock $1 sh -c "$2"; \
 	elif command -v lockf >/dev/null; then \
-		lockf $1 sh -c "$2"; \
+		lockf -k $1 sh -c "$2"; \
 	else \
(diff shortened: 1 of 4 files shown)


Zerpet:
core-autoquery is fixed. core-autopatch is still failing. Marking as draft to investigate further.

essen:
You probably also want to add macos/windows as they are currently disabled in the workflow. Not sure the workflow will work for them though.

Zerpet:
PR is green! 🙌  I can work on enabling MacOS runners next, but I'd like to do that in a different PR.

essen:
Merged, thanks!

PR #1047, opened 2026-05-26
Use `.test` domain for test emails

Zerpet:
Following the convention for test emails. Reference RFC 2606. It is guaranteed that `.test` top level domain will never be assigned to someone on the Internet.


essen:

